PR #2, opened 2019-05-06
improve README format: new line for QR code

oldratlee:


li-qing:
newline before QRCode is added

PR #14, opened 2019-05-07
Fix dimension type

daquexian:
The return value of `Tensor::DimensionType Tensor::getDimensionType() const` is wrong and inconsistent with `static Tensor::DimensionType getDimType(const Tensor* origin)`

### DIFF
--- a/source/core/Tensor.cpp
+++ b/source/core/Tensor.cpp
@@ -162,9 +162,8 @@ bool Tensor::copyToHostTensor(Tensor* hostTensor) const {
     return true;
 }
 
-static Tensor::DimensionType getDimType(const Tensor* origin) {
-    auto dimformat = TensorUtils::getDescribe(origin)->dimensionFormat;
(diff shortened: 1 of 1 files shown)


li-qing:
getDimType is a private function in tensor.cpp, and it's implemented intentionally in this way to simplify dimension type handle logic.

daquexian:
> getDimType is a private function in tensor.cpp, and it's implemented intentionally in this way to simplify dimension type handle logic.

So is it an intended behavior that `Tensor::DimensionType Tensor::getDimensionType() const` returns CAFFE even if the right type is CAFFE_C4?

jxt1234:
"So is it an intended behavior that Tensor::DimensionType Tensor::getDimensionType() const returns CAFFE even if the right type is CAFFE_C4?"  Yes.

This is a legacy problem. Before the modification, we need to correct the relevant code.

li-qing:
yes, so far.

PR #93, opened 2019-05-23
Add Upsample and Constant op type support for onnx converter

Howave:


czy2014hust:
Appreciate your contribution! 
The initializers of the ONNX model should not be modified, The ONNX Constant node should be converted to MNN Const op.

Howave:
How about passing the constant node attributeProto to the opInitializers of Upsample op?

tools/converter/source/onnx/ConstantOnnx.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+    const onnx::TensorProto *constantTp;
+    for (int i = 0; i < onnxNode->attribute_size(); ++i) {
+        const auto &attributeProto = onnxNode->attribute(i);
+        const auto &attributeName  = attributeProto.name();
+        if (attributeName == "value") {
+            constantTp = &attributeProto.t();
+        }
+    }
+
```
czy2014hust, line 35: Here, Check constantTp is not nullptr

Howave:
Would pls have a check again?

PR #155, opened 2019-07-11
Fix: bug in int8 convolution when dilateX is not equal to dilateY

yisongsong:
Fix: bug in int8 convolution when dilateX is not equal to dilateY


suitmyself:
It seems that should also change code at line 150 from
`TensorUtils::setLinearLayout(&mTempBuffer);`
to
`TensorUtils::setLinearLayout(&mSrcCopyBuffer);`

PR #192, opened 2019-07-26
Fix tf lrn wrong localsize

daquexian:
Reference: https://www.tensorflow.org/api_docs/python/tf/nn/local_response_normalization

tf lrn `depth_radius` parameter indicates the _radius_ rather than the size of channels, so the size should be `2 * depth_radius + 1`


daquexian:
我在我的 [DNNLibrary](https://github.com/JDAI-CV/DNNLibrary/blob/master/tools/onnx2daq/OnnxConverter.cpp#L1497-L1516) 里也踩过这个坑

PR #226, opened 2019-08-11
fix compareTensors() function

korolevivan:


leo-nullptr:
@korolevivan You should use fabsf for `float` type.

korolevivan:
> @korolevivan You should use `fabsf` for `float` type.

thanks, fixed)

li-qing:
sign the CLA plz so that I can merge your pull request

gdh1995:
Then I think the line 132 should check `!std::isinf(pb[i])` first. Am I right?

PR #315, opened 2019-09-12
avoid protobuf error

twmht:
With latest protobuf 3.9.1, MNNConvert would throw the errors if not compiled with pthread

```
[libprotobuf FATAL google/protobuf/generated_message_util.cc:783] CHECK failed: (scc->visit_status.load(std::memory_order_relaxed)) == (SCCInfoBase::kRunning): 

terminate called after throwing an instance of 'google::protobuf::FatalException'
  what():  CHECK failed: (scc->visit_status.load(std::memory_order_relaxed)) == (SCCInfoBase::kRunning): 
```

### DIFF
--- a/tools/converter/CMakeLists.txt
+++ b/tools/converter/CMakeLists.txt
@@ -99,4 +99,4 @@ else()
     target_link_libraries(MNNConvert tensorflow caffe onnx tflite)
 endif()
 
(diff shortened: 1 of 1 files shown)


li-qing:
the 'pthread' problem have been fixed in recent releases. so close this pull request. thanks for your pull request all the same.

PR #440, opened 2019-11-05
Fix pydot error

gasgallo:
mnnvisual used to throw error due to pydot unable to handle bytestrings, I added `decode()` to solve this issue.

### DIFF
--- a/pymnn/pip_package/MNNTools/mnnvisual.py
+++ b/pymnn/pip_package/MNNTools/mnnvisual.py
@@ -95,7 +95,7 @@ def mnn_to_dot(mnn_file):
     for idx in range(ts_num):
         ts_name = net.TensorName(idx)
         name = "tensor_" + str(idx)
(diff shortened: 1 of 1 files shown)


CLIsVeryOK:
I met the following error using mnnvisual, anyone helps?
pydot_ng.InvocationException: Program terminated with status: 1. stderr follows: Error: /tmp/tmp2OO22j: syntax error in line 1472 near ':'

stanleywang8888:
> I met the following error using mnnvisual, anyone helps?
> pydot_ng.InvocationException: Program terminated with status: 1. stderr follows: Error: /tmp/tmp2OO22j: syntax error in line 1472 near ':'

two cases:
case1 : local pydot is not correctly installed, you should apt-get install pydot,  not only pip install pydot python package 
case2 :  probably you're using a very old python package, later versions like 1.0.0 and 1.0.1 already fixed the issue

PR #511, opened 2019-12-10
Fixed get_model.sh script unable to download model

JackyAndroid:
Fixed a problem where the get_model.sh script could not download the model if the build folder did not exist


Naville:
We expect our user building MNN under ${SRC_ROOT}/build , which is also the path we look for the converter binary. Thus I don't think I understand the purpose of this patch, or am I missing something here?

PR #687, opened 2020-03-17
fix spelling mistake, eltwies -> eltwise

Guanmoyu:
1、fix spelling mistake, eltwies -> eltwise;
2、I executed the script, generate.sh；
3、I don't have the environment of metal, so CPUOPRegister.cpp is not updated.


Guanmoyu:
Sorry, I have removed the schema/current. In the file .gitignore, should the "schema/current" be uncommented?

PR #824, opened 2020-05-24
use "find" instead of iteration

gkUwen:
Hi,
 
For unordered_map,   I think that "find" would be more efficient than iteration. 
Or I ignore some reasons so that we use iteration? Please check it.


jxt1234:
You are right. Please signed the CLA so we can merge it.

PR #848, opened 2020-05-31
Add BlstmComputer

Interfish:
Main changes
1.  Add blstmcomputer
2. Add blstmcomputer unit test
3. add matrix dot product method to math/matrix.hpp
4. modify CmakeLists under /test a little to pass test building 

At first I was tring to convert Tensorflow's op 'CudnnRNNV3' into MNN op, but later on I found that kernel weight in 'CudnnRNNV3' op isn't static and the shape may change between different cudnn versions. Thus there is no easy and stable way to convert that op.
When I realized this fact, I had already finished developing this blstmcomputer. So I open this pull request and hope my work can benefit to other's. Anyone who wish to use blstm in their model can reuse this blstmcomputer to greatly speed up their developing process. All unit tests are passed and I believe this blstmcomputer always yields correct outcome.


source/backend/cpu/compute/BlstmComputer.cpp
```diff
@@ -0,0 +1,346 @@
...
+      backend()->onAcquireBuffer(mGateOutputs[i].get(), Backend::DYNAMIC);
+    }
+
+    mOutput.reset(Tensor::createDevice<float>(
+        vector<int>{batchSize * timeSteps,
+                    mBidirectional ? 2 * mStateSize : mStateSize},
+        Tensor::CAFFE));
+    backend()->onAcquireBuffer(mOutput.get(), Backend::DYNAMIC);
+  }
+  mBatchSize = batchSize;
```
jxt1234, line 184: After the mid tensor usage, it should be released by onReleaseBuffer

Interfish: Thanks, my fault. Already resolved

jxt1234: Thanks.

PR #923, opened 2020-06-24
BugFix: converter optimizer forget to copy 「outputName」

chrisYooh:
Problem Inference:
1. The json file from DumpMNN2json will not include outputName info
2. Inference model.mnn will not enable to get multi-outputs
3. Inference model.mnn perhaps get wrong output.


jxt1234:
Thanks for contribute, please signed the CLA and we can merge it.

chrisYooh:
CLA signed done.

PR #945, opened 2020-07-02
Don't increase reference count when extracting a block from a non-splitable freelist

krayzemli:
... since returning a block to a non-mergeable free list does not decrement this count.

Looks like a bug for me that could lead to the leak of free memory blocks. Please verify if this is really so. Found when looking for a different memory bug.


hush-alibaba:
@krayzemli Hello, MNN team is in the process of building a group of core contributors to the project. I saw you made a few pull requests in the past few weeks. Would you be interested in it?

krayzemli:
That is possible. Although my interest in MNN framework is episodic. What benefits and responsibilities would your status of a core contributor give to me?

PR #991, opened 2020-07-22
Improved performance in DirectedAcyclicGraph. 

proydakov:
Using `find` instead of `foreach` in std::unordered_map.

### DIFF
--- a/source/core/DirectedAcyclicGraph.hpp
+++ b/source/core/DirectedAcyclicGraph.hpp
@@ -174,20 +174,16 @@ private:
                 const weak_ptr<Node<T> > oNode = outEdge->getDst();
                 if (!oNode.expired()) {
                     const shared_ptr<Node<T> > node = oNode.lock();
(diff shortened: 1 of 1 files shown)


source/core/DirectedAcyclicGraph.hpp
```diff
@@ -174,20 +174,15 @@ class DirectedAcyclicGraph {
...
-                                TopoNode tn;
-                                tn.node     = node;
-                                tn.outEdges = node->getOutEdges();
-                                noIncoming.push_back(tn);
-                                nodesInEdges.erase(edg_iter);
-                            }
-                            break;
+                    auto edg_iter = nodesInEdges.find(node);
+                    if (edg_iter != nodesInEdges.end()) {
+                        if (edg_iter->second.size() <= 0) {
```
jxt1234, line 181: Left edg_iter->second.erase(outEdge)

proydakov: Fixed. thx @jxt1234

PR #993, opened 2020-07-23
Feature/sync

jxt1234:


jxt1234:
Fix bug for  opencl and pymnn.
Support fp16 compress for matmul

PR #1049, opened 2020-08-12
TFLITE PRelu support and fixing vulkan validation errors

rinthel:
Hello, I'm a developer living in Seoul, Korea.

I'm a newbie of the MNN, and first would like to say thank you for developing such a great framework.
I'm interested in the inference engine with Vulkan acceleration on a desktop application,
so I tried to compile the MNN project on my Windows machine,
and found some limitations and errors as below:
- TFLITE converter does not support the PRelu operation, even though MNN itself already supports it.
- The Padding operator for Vulkan backend is not implemented, thus it goes to CPU backend, which can make the session slower.
- Some Vulkan validation errors, especially due to image layout setting.

This PR is a draft for fixing above. Here is what I tried:
- MNNConverter
  * `prototxt` parameter does not have to be given when we don't use it for the CAFFE framework
  * Add PRelu converter implementation for TFLITE framework
- Vulkan backend: bug
  * In `VulkanBackend::findTensor()`, it also look up in `mStaticBuffers` as well as `mAllBuffers`.
    This modification can be fix crash bug when the output tensor set as a static buffer.
  * Add `VulkanPadding`, which is a simple image copy for zero-padding. It should be revised
- Vulkan backend: validation error
  * In buffer mapping and flushing process, align size in according to the physical device's limits 
  * Add some missing `sType` flags
  * Add `VK_DESCRIPTOR_POOL_CREATE_FREE_DESCRIPTOR_SET_BIT` flag for creating descriptor pool in order to notify that we use `vkFreeDescriptorSet` function
  * Add a `mLayout` property in `VulkanImage`, which is used to track its layout type
  * Add a `barrierImageIfNeeded()` method in `VulkanCommandPool::Buffer`, which is a utility function to set the image layout when it's necessary.
  * Add additional parameter `finalLayout` for `VulkanBackend::copyBufferToImage()` to set image layout after finishing copy behaviour.
  * Switch `barrierImage()` to `barrierImageIfNeeded()`
  * Use `barrierImageIfNeeded()` for input and output tensors before dispatching.

I've tested the code in the Windows 10 / AMD Ryzen 3900X / nVidia GeForce 2080Ti,
and tested `expressDemo` with `deeplabv3`, `squeezenet-v1.1`, and `mobilenet-v2.1.tflite` networks.

Please let me know for any suggestion.
Thank you for reading!

### DIFF
--- a/source/backend/vulkan/execution/VulkanMatrixMultier4x4.cpp
+++ b/source/backend/vulkan/execution/VulkanMatrixMultier4x4.cpp
@@ -100,8 +100,10 @@ void VulkanMatrixMultier4x4::prepare(int e, std::shared_ptr<VulkanImage> dst, st
 
 void VulkanMatrixMultier4x4::compute(const VulkanCommandPool::Buffer* commandBuffer) const {
     mPipeline->bind(commandBuffer->get(), mDescriptorSet->get());
-    commandBuffer->barrierImage(mSource->get(), VK_IMAGE_LAYOUT_GENERAL, VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL);
-    commandBuffer->barrierImage(mKernel->get(), VK_IMAGE_LAYOUT_GENERAL, VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL);
+    commandBuffer->barrierImageIfNeeded(mSource.get(), VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL);
+    commandBuffer->barrierImageIfNeeded(mKernel.get(), VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL);
+    // commandBuffer->barrierImage(mSource->get(), VK_IMAGE_LAYOUT_GENERAL, VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL);
+    // commandBuffer->barrierImage(mKernel->get(), VK_IMAGE_LAYOUT_GENERAL, VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL);
     vkCmdDispatch(commandBuffer->get(), UP_DIV(mOutputWidth, 8), UP_DIV(mOutputHeight / 4, 8), mDepth);
 }
 } // namespace MNN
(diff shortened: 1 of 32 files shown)


jxt1234:
Thanks for your pull request. Could you signed the CLA and then I can merge it.

source/backend/vulkan/execution/VulkanMatrixMultier4x4.cpp
```diff
@@ -82,6 +82,10 @@ void VulkanMatrixMultier4x4::prepare(int e, std::shared_ptr<VulkanImage> dst, st
         mDest->release();
     }
 
+    commandBuffer->barrierImageIfNeeded(mDest.get(), VK_IMAGE_LAYOUT_GENERAL);
+    commandBuffer->barrierImageIfNeeded(mSource.get(), VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL);
+    commandBuffer->barrierImageIfNeeded(mKernel.get(), VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL);
```
jxt1234, line 87: prepare and compute both barrier may be not needed?

rinthel: @jxt1234 You are right, my code was quite defensive. Maybe we can remove barrier-related code in `compute()`

rinthel:
```c++
void VulkanMatrixMultier4x4::compute(const VulkanCommandPool::Buffer* commandBuffer) const {
    mPipeline->bind(commandBuffer->get(), mDescriptorSet->get());
    vkCmdDispatch(commandBuffer->get(), UP_DIV(mOutputWidth, 8), UP_DIV(mOutputHeight / 4, 8), mDepth);
}
```
I've checked that this modified code also does not cause validation errors. I'll push it.

121786404:
https://raw.githubusercontent.com/aqua1907/Gesture-Recognition/master/models/palm_detection_without_custom_op.tflite

use  https://github.com/rinthel/MNN/tree/tflite-conversion-and-vulkan-validation

/MNN/tools/converter/source/tflite/ReshapeTflite.cpp:55: Check failed: shapeTensor->type == tflite::TensorType_INT32 ==> ERROR

rinthel:
@121786404 Sorry, but I think that the error is not related to this PR, because I did not modify `ReshapeTflite.cpp` at all. Did you check that the error occurs when you use the master branch of MNN? If so, it should be treated as a new issue.

121786404:
> @121786404 Sorry, but I think that the error is not related to this PR, because I did not modify `ReshapeTflite.cpp` at all. Did you check that the error occurs when you use the master branch of MNN? If so, it should be treated as a new issue.

master branch of MNN not support PRelu，will error when convert PRelu

PR #1202, opened 2020-10-30
fix MNNConvert did not output fatal errors

gdh1995:
If `LOG(FATAL)` and `CHECK(...)` throws a `std::runtime_error` instance,
the `main()` doesn't catch it, so no message can be seen (tested on cmd of Win 10 19041).


jxt1234:
Please update and solve the confict and then we can merge it.

gdh1995:
Hello. The conflict has been fixed.

PR #1246, opened 2020-11-19
Modern CPP cleanup

shouwei:
this pull request addresses the following issues:
1. exclude std namespace from the header file.
2. replace C-style with modern C++, such as static_cast, const_cast, etc.
3. correct typos.
4. some cleanup.
5. std::move() has no effect on the const reference objects.
6. prefer ++it to it++.
7. prefer emplace() to insert().
8. replace the include guard with #pragma once.
9. strongly typed enum.
10. not a good idea to mix use smart pointer and raw pointer, particularly for the following case:
```cpp
    class A;
    A* func() {
        std::shared_ptr<A> sp = std::make_shared<A>();
        return sp.get();
    }
```
sp is a local object on the stack. When func() returns, the object will be destroyed and the accompanying memory will be released. Thus, sp.get() probably returns a wild pointer.


include/MNN/Interpreter.hpp
```diff
@@ -249,16 +248,16 @@ class MNN_PUBLIC Interpreter {
...
-    Tensor* getSessionInput(const Session* session, const char* name);
+    Tensor* getSessionInput(const Session* session, const char* name) const;
     /**
      * @brief get output tensor for given name.
      * @param session   given session.
      * @param name      given name. if NULL, return first output.
      * @return tensor if found, NULL otherwise.
      */
-    Tensor* getSessionOutput(const Session* session, const char* name);
+    Tensor* getSessionOutput(const Session* session, const char* name) const;
```
jxt1234: We can't modify old api currently. Because some app will only update mnn.so but not compile their code with modified head file.

shouwei: I already reverted all interface changes in "Interpreter.hpp". Since we use MNN in our product, I prefer to commit changes here with PR to avoid unnecessary conflicts with our cloned repository.

PR #1271, opened 2020-12-03
fix threadpool may not destroyed

DaydreamCoding:


jxt1234 (approved):

gdh1995:
Could you explain why the old `mStop: atomic_bool = true` is not safe, and why a lock surrounding it is enough? Sorry I'm not good at multi-thread programming (I'm learning it by reading MNN).

PR #1298, opened 2020-12-15
GeometryInnerProduct: embed bias to MatMul if batch size is 1

gdh1995:
This should be a little faster.

### DIFF
--- a/source/geometry/GeometryInnerProduct.cpp
+++ b/source/geometry/GeometryInnerProduct.cpp
@@ -79,7 +79,7 @@ public:
         
         std::shared_ptr<Tensor> tmpOutput;
         std::shared_ptr<Tensor> C(new Tensor);
-        auto constTensors = context.searchConst(op);
+        auto& constTensors = context.searchConst(op);
         Tensor* weight = nullptr;
         Tensor* bias = nullptr;
         if (!constTensors.empty()) {
--- a/source/geometry/GeometryInnerProduct.cpp
+++ b/source/geometry/GeometryInnerProduct.cpp
@@ -22,6 +22,7 @@ public:
         auto parameter  = op->main_as_InnerProduct();
         int outputCount = parameter->outputCount();
         int srcCount    = parameter->weight()->size() / outputCount;
+        bool hasBias    = parameter->biasTerm() > 0;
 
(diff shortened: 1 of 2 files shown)


source/geometry/GeometryInnerProduct.cpp
```diff
@@ -22,6 +22,7 @@ class GeometryInnerProduct : public GeometryComputer {
         auto parameter  = op->main_as_InnerProduct();
         int outputCount = parameter->outputCount();
         int srcCount    = parameter->weight()->size() / outputCount;
+        bool hasBias    = parameter->biasTerm() > 0;
```
jxt1234, line 25: it may cause old model result error. Currently the innerproduct is not used for future.

gdh1995: Up to now, only caffe and onnx converter will set this flag:
* onnx converter always sets it `1` in `tools/converter/source/onnx/MatMulOnnx.cpp`
* caffe converter uses `par.bias_term()`. If it's not 1 but 0, then converter will use an array filled with `0`

So converted models will not be affected by this PR.

BTW, the official code even converts caffe's InnerProduct to Conv 1x1, so ... While I commented the line to get MNN models with `OpType_InnerProduct`, so that I can debug such layers easier.

gdh1995: Well, it seems `biasTerm` is default to 0 in the `.fbs` file. Then this line may be replaced with `bool hasBias = true`, while other checks like `batch > 1` can be kept.

gdh1995: Thanks for this comment. I've noticed and confirmed that this PR causes a conflict:
* `GeometryComputerUtils::makeMatMul` always gets a valid `bias`
* if `batch > 1`, another `makeBinary` also gets the `bias`

Then when `batch > 1`, this PR causes result error...

This conflict can be fixed by removing `makeBinary` completely, but there's still another issue: `makeMatMul` uses a fast path when `batch == 1`, which makes the result of `A*X+B` a bit different from `A * [ X[0], X[0] ] + B` if `A` is `1xCxHxW`.

Overall, this PR is not so good, and does deserve the reverting ...

PR #1395, opened 2021-03-03
fix opencl runtime error of MSVC; add 'clGetMemObjectInfo' wrapper

WillTao-RD:
1. fix opencl runtime error of MSVC

Due to a BUG of VS compiler which makes 'call_once' function executing twice and is not fixed yet([link](https://developercommunity.visualstudio.com/t/dynamic-initializer-created-for-stdonce-flag/243273)), 
avoid to use global static variable, and use local static variable instead.

2. add 'clGetMemObjectInfo' wrapper

### DIFF
--- a/source/core/Backend.cpp
+++ b/source/core/Backend.cpp
@@ -17,8 +17,8 @@ namespace MNN {
 
 void registerBackend();
 
-static std::once_flag gInitFlag;
 static std::map<MNNForwardType, std::pair<const RuntimeCreator*, bool>>& GetExtraCreator() {
+    static std::once_flag gInitFlag;
     static std::map<MNNForwardType, std::pair<const RuntimeCreator*, bool>>* gExtraCreator;
     std::call_once(gInitFlag,
                    [&]() { gExtraCreator = new std::map<MNNForwardType, std::pair<const RuntimeCreator*, bool>>; });
--- a/source/backend/opencl/core/runtime/OpenCLWrapper.cpp
+++ b/source/backend/opencl/core/runtime/OpenCLWrapper.cpp
@@ -164,6 +164,7 @@ bool OpenCLSymbols::LoadLibraryFromPath(const std::string &library_path) {
     MNN_LOAD_FUNCTION_PTR(clGetKernelWorkGroupInfo);
     MNN_LOAD_FUNCTION_PTR(clGetEventInfo);
     MNN_LOAD_FUNCTION_PTR(clGetEventProfilingInfo);
+    MNN_LOAD_FUNCTION_PTR(clGetMemObjectInfo);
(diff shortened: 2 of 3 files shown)


source/core/Backend.cpp
```diff
@@ -17,8 +17,8 @@ namespace MNN {
 
 void registerBackend();
 
-static std::once_flag gInitFlag;
 static std::map<MNNForwardType, std::pair<const RuntimeCreator*, bool>>& GetExtraCreator() {
+    static std::once_flag gInitFlag;
     static std::map<MNNForwardType, std::pair<const RuntimeCreator*, bool>>* gExtraCreator;
```
jxt1234, line 22: Why move gInitFlag inside?

WillTao-RD: Sorry for the late reply.

Setting `gInitFlag` outside as a global variable is absolutely RIGHT. 
Putting it inside is just to make it **compatible with MSVC**.

### REASON
MSVC compiler has a [BUG](https://developercommunity.visualstudio.com/t/dynamic-initializer-created-for-stdonce-flag/243273) that will treat constructor of once_flags as `dynamic initializer`, which should be `constant initialier`, and this bug has not be fixed till now.

### IF NOT DO
Some backends are registed in the background by static variable at early runtime and will be erased later. 

Take opencl as an example.

https://github.com/alibaba/MNN/blob/1c4eded0148af60b3c07da64dcb6a79358f827b2/source/backend/opencl/core/OpenCLBackend.cpp#L679-L681
https://github.com/alibaba/MNN/blob/1c4eded0148af60b3c07da64dcb6a79358f827b2/source/core/Backend.cpp#L48-L50
https://github.com/alibaba/MNN/blob/1c4eded0148af60b3c07da64dcb6a79358f827b2/source/core/Backend.cpp#L20-L26

1. In the code above. `gResistor` calls the `std::call_once` creator to new an opencl-backend for `gExtraCreator`, and set `gInitFlag` to **NOT ZERO**.

2. Then `gInitFlag` is constructed by `dynamic initializer`, and it will be set to **ZERO** again. 

3. In the code below, when `registerBackend` function is called by the class constructor of `Executor`, `GetExtraCreator` calls `std::call_once` again.

https://github.com/alibaba/MNN/blob/1c4eded0148af60b3c07da64dcb6a79358f827b2/express/Executor.cpp#L101-L103
https://github.com/alibaba/MNN/blob/1c4eded0148af60b3c07da64dcb6a79358f827b2/source/core/Backend.cpp#L28-L31

4. At this time `gInitFlag` is still zero, so `std::call_once` will be executed for the second time, which will create another map for `gExtraCreator`. The backends registed before are clear and memory leaks.

jxt1234 (approved):

PR #1435, opened 2021-04-07
add return value for setCache, fix cache update bug

WillTao-RD:
## Issue Describe
If there is an old cache file which is built by an old model, and now we are trying to create a new session using the cache file.
It reports error everytime but the cache file is not updated at all.

## Bug Fix
1. Currently, cache is rebuilt and saved when `if ((!mNet->cacheFile.empty()) && (!valid)) ` is true.
However, **`valid` indicates whether old cache is loaded successfully or not**. If load failed, rebuild and save are STILL needed. so it should be `if ((!mNet->cacheFile.empty()) || (!valid))`.
2. At the same time, cache should be reset before rebuild and save.
3. In addition, `setCache` should return the right value to ensure `valid` is reliable.


jxt1234:
Thanks for contribute. Please fix conflicts and I can merge it.

WillTao-RD:
Sorry for the late reply

PR #1599, opened 2021-07-30
[bugfix] fix build static library with ARMV82

DaydreamCoding:


CMakeLists.txt
```diff
@@ -322,6 +322,9 @@ list(APPEND MNN_TARGETS MNNCore)
 if(MNN_BUILD_MINI)
     target_compile_options(MNNCore PRIVATE -DMNN_BUILD_MINI)
 endif()
+if(MNN_ARM82)
+    target_compile_options(MNNCore PRIVATE -DENABLE_ARMV82)
```
jxt1234, line 326: It may cause dup registerArm82RuntimeCreator when build for shared library.
Need modify source/backend/arm82/Arm82Backend.cpp and remove __arm82_global_initializer

DaydreamCoding: ENABLE_ARMV82 is private here : https://github.com/alibaba/MNN/blob/master/source/backend/arm82/CMakeLists.txt#L11

but the target core do not get ENABLE_ARMV82, with static library can not fall into
 https://github.com/alibaba/MNN/blob/master/source/core/BackendRegister.cpp#L42

Do you have try static library with armv8.2?

PR #1640, opened 2021-08-24
fix bug of header file missing when build for aarch64 with open ARM82

jiuzhuanzhuan:


jxt1234:
Thanks for pull request, please signed the CLA and then We can merge it.

PR #1746, opened 2021-11-04
add CPUROIAlign op and unit-test and so on

no5-aaron-wu:
1. add ROIAlignOnnx convert
2. add new test case in ROIPoolingTest.cpp
3. add CPUROIAlign op and uint test
4. add neon realization of CPUROIAlign op


no5-aaron-wu:
i found some problems, fixing now, no rash to merge

schema/current/MNN_generated.h
```diff
@@ -247,11 +247,12 @@ enum OpType {
   OpType_If = 601,
   OpType_LayerNorm = 603,
   OpType_GridSample = 604,
+  OpType_ROIAlign = 605,
```
jxt1234: New Op type need be added before OpType_Plugin, Better use.OpType_ROIAlign = 143

schema/default/UserDefine.fbs
```diff
@@ -14,9 +14,22 @@ enum BorderMode : byte {
...
 
 table GridSample {
     mode:SampleMode;
     paddingMode:BorderMode;
     alignCorners:bool=false;
 }
+
+table RoiAlign {
+    pooledWidth:int;
+    pooledHeight:int;
```
jxt1234: RoiAlign Parameter Can be modified from RoiPooling. Just add other parameter to table RoiPooling

no5-aaron-wu: I know they have some parameters in common, but it's strange to use `RoiPoolingT` directly in ROIAlign op.It's hard to understand at first look. or will be better using the RoiPooling table as a parameter in RoiAlign like follow?
```
table RoiAlign {
    roiPoolingParam:RoiPooling;
    samplingRatio:int=-1;
    aligned:bool=false;
    poolMode:PoolMode=AvePool;
}
```

schema/current/MNN_generated.h
no5-aaron-wu: get it

schema/default/UserDefine.fbs
jxt1234: You can rename RoiPooling to RoiParameters

no5-aaron-wu: 1

no5-aaron-wu:
review problems fixed. pls check if exist other problem or not.

jxt1234 (approved):

PR #1799, opened 2021-12-23
fix android demo load OpenCL library

mirsfang:


jxt1234:
Please signed CLA and then I can merge it.

PR #1937, opened 2022-05-27
Fix some typos in 3rd_party.

Yulv-git:
Fix some typos in 3rd_party.


jxt1234 (commented):
3rd party is not written by MNN. It may be overwritten after sync. Normally don't change it.

Yulv-git:
> 3rd party is not written by MNN. It may be overwritten after sync. Normally don't change it.

Ok, thanks.

PR #1988, opened 2022-07-12
Add external data support for onnx converter

brian14708:


jxt1234 (approved):
Test for inside onnx models and pass.

PR #2047, opened 2022-09-07
Add Interpolation 3D (Interp3d)

wtiandong:
1. Add Interpolation 3D for ONNX Converter
2. Add Interp3D Nearest CPU Implementation.
3. Add Interp3D Nearest Opencl Implementation.

Interp3D operators (high precision) are verified with PyTorch implementation with np.allclose


wtiandong:
wrong name

PR #2048, opened 2022-09-07
Add Interp3D Support

wtiandong:
1. add PyTorch interpolation 3D to Onnx to MNN converter
2. add interpolation3D nearest CPU/OpenCL implementation
3. update opencl_program.cc
all added OPs are verified


schema/default/MNN.fbs
```diff
@@ -138,6 +138,7 @@ enum OpType : int {
     OneHot,
     BroadcastTo,
     Dilation2D,
+    Interp3D,
```
jxt1234, line 141: Interp3D OP can be fused into Interp.

wtiandong: I think a separate implementation is better. Just as Conv and Conv3D - different Op_Params, implementation APIs, etc. Moreover, due to the change of Op_params of Interp, compatibility problems will also lead to lots of obstacles to thousands of existing models, they need to re-convert due to the update of "Interp". Meanwhile, maybe more than 98% of MNN models are just for 2D images, not for 3D volumes, why do 98% of MNN models need to re-covert because of the  2% 3D volume demands?

> Interp3D OP can be fused into Interp.

wtiandong: As discussed in DingTalk, I merged interp3D Param into Interp Param

jxt1234: OK

jxt1234 (approved):

PR #2063, opened 2022-09-15
Pipeline command cache

DaydreamCoding:
optimize performance

### DIFF
--- a/source/core/Pipeline.cpp
+++ b/source/core/Pipeline.cpp
@@ -925,6 +925,16 @@ ErrorCode Pipeline::allocMemory(bool firstMalloc) {
         }
     }
     mBackend->onResizeEnd();
+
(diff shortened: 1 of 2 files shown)


source/core/Pipeline.cpp
```diff
@@ -296,6 +296,15 @@ ErrorCode Pipeline::encode(bool isStatic, bool supportDebug) {
     }
 #endif
 
+    // cached
+    command_cached.clear();
+    for (auto& info : mInfo) {
```
jxt1234: Merge command in info is to reduce the compute time for Geometry::encode.

DaydreamCoding: in some model, mInfo->executeBuffer->command frequently empty

PR #3079, opened 2024-11-14
Vulkan uses proper CMake Vulkan target

EduMenges:
Current setup uses the raw path, which is problematic when building a project that uses MNN and offers a `Config` file, because the absolute path of Vulkan is embedded into MNN's `Config` file. This makes use of [the proper CMake target](https://cmake.org/cmake/help/latest/module/FindVulkan.html), `Vulkan::Vulkan`.

I had to bump CMake's minimum version to 3.24 because the `GLOBAL` attribute required for this fix is only available from that version. Version 3.6 shouldn't be used anyway, as versions < 3.10 [were deprecated in CMake 3.31](https://cmake.org/cmake/help/latest/release/3.31.html#deprecated-and-removed-features).


CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.6)
+cmake_minimum_required(VERSION 3.24)
```
jxt1234, line 1: It seems cmake version update too large.

PR #3101, opened 2024-11-25
Bump MNN KleidiAI ukernel to qai8_qsi4_sme2 ukernel

xhzheng1895:
Add logic to select ukernels based on modelInfo and CPUInfo in mnn_kleidiai.cpp. Move some pack functions to mnn_kleidiai_util.cpp.

Add CPU feature detect in source/backend/cpu/CPURuntime.hpp.

Thread number will be forced to 1 when SME2 is enabled, for better energy efficiency ratio.


xhzheng1895:
Interface changed. Close this pr and create a new one later.

PR #3149, opened 2025-01-03
fix: MSVC SIMD compiling

EduMenges:
I have found issues when compiling the SIMD targets for MSVC.

Firstly, the target `MNNAVX` requires the AVX2 instruction set. While this is correctly set when using Clang:

```cmake
target_compile_options(MNNAVX PRIVATE -mavx2 -DMNN_X86_USE_ASM)
```

It is not correct when using MSVC, which is still set to `AVX`. This prevents correct compiling.

The same is true for the `MNNSSE` target, which requires the `AVX` arch (the `SSE` arch is ignored in x64).

Lastly, there is a weird 256 bit dereference as if it is a struct, but in MSVC it is an opaque type, thus it should be dereferenced using SIMD intrinsics. I've updated the code so that the GCC/MSVC distinction is not necessary.

The fact that MNN's CI for Windows does not catch these errors makes me think the SIMD parts are not getting compiled during CI.


EduMenges:
> What's your version of Vision Studio. It seems not meet compile error for 2017 and later.

It's the 2022 Community Edition, more precisely version 17.12.3.

PR #3212, opened 2025-02-10
Optimize Stft func

linlinxu-1989:
1. When the CPUStft constructor is called for the first time, initialize gCosTable and gSinTable.
2. When calling the CPUStft execution function, retrieve the corresponding values from gCosTable and gSinTable based on the index.


jxt1234:
The CPUSTFT will be rewrite in next version, but the pull request can merge. Please sign CLA

PR #3219, opened 2025-02-11
Migrate the android Vision transformer demo in mnn-llm 

HenryDen:
Migrate the android Vision transformer demo from mnn-llm to MNN/transformer/llm/engine.
Changes:
1. Change the UI style and add MNN logo and kleidiAI logo.
2. Pack the model in the apk, the model will be download at android/app/src/main/assets/models
3. Require for all file access right to avoid read picture failure.


source/core/MNNFileUtils.h
```diff
@@ -22,6 +22,7 @@
 #undef NO_ERROR
 #else
 #include <unistd.h>
+#include <errno.h>
```
jxt1234: this line can delete

HenryDen: Deleted in the latest commit

PR #3403, opened 2025-04-28
[LLM:Bugfix] Fix incorrect kv_seq_len in gen_attention_mask() for autoregressive decoding

HirokenOvo:
During the **decoding** phase (auto-regressive generation), `Llm::generate()` calls `forward(input_ids, is_prefill)` with `input_ids.size() == 1`. The current logic in `Llm::gen_attention_mask()` sets `kv_seq_len = seq_len` (where `seq_len=1` for decoding), which is incorrect.  This ignores the accumulated history length（`mContext->all_seq_len`）, leading to mismatched attention masks and affecting the calculation of attention scores, thereby influencing the model's inference results.

### DIFF
--- a/transformers/llm/engine/src/llm.cpp
+++ b/transformers/llm/engine/src/llm.cpp
@@ -720,9 +720,7 @@ std::string Llm::tokenizer_decode(int id) {
 
 VARP Llm::gen_attention_mask(int seq_len) {
     int kv_seq_len = mContext->all_seq_len + seq_len;
-    if (seq_len == 1) {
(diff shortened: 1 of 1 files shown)


HirokenOvo:
Noticed that [commit `bd36a3f`](https://github.com/alibaba/MNN/blob/bd36a3f749836531aa19903b8575d7ee91565aff/transformers/llm/engine/src/llm.cpp) has addressed the issue of incorrect `kv_seq_len` calculation in `gen_attention_mask()`, this PR is no longer necessary,  I’ll close this PR now.

PR #3569, opened 2025-05-27
Bump MNN KleidiAI ukernel to fp16_sme2 fp32_sme2 ukernel

linlinxu-1989:
Add fp16 and fp32 sme2 kernel
1. Add rhs pack nxk kernel lhs pack kernel
2. Add fp16 GEMM kernel and GEMV kernel
3. Add fp32 GEMM kernel and GEMV kernel

### DIFF
--- a/source/backend/cpu/compute/Convolution1x1Strassen.hpp
+++ b/source/backend/cpu/compute/Convolution1x1Strassen.hpp
@@ -26,7 +26,9 @@ public:
     virtual bool onClone(Backend* bn, const Op* op, Execution** dst) override;
 private:
     std::shared_ptr<CPUConvolution::Resource> mResource;
+#ifdef MNN_KLEIDIAI_ENABLED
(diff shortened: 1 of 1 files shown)


source/backend/cpu/compute/Convolution1x1Strassen.hpp
```diff
@@ -26,6 +26,7 @@ class Convolution1x1Strassen : public CPUConvolution {
     virtual bool onClone(Backend* bn, const Op* op, Execution** dst) override;
 private:
     std::shared_ptr<CPUConvolution::Resource> mResource;
+    std::shared_ptr<Tensor> mInputResource;
```
jxt1234, line 30: need MNN_KLEIDIAI_ENABLED macro wrap

linlinxu-1989: ok

PR #3581, opened 2025-05-29
Fix: remove automatic remote repository check at startup

ghost:
#### Problem
The application currently performs an automatic remote repository check during startup in the ModelListPresenter class. This causes the app to get stuck on the loading screen indefinitely when there's no internet connection or when the remote model repository is unreachable.
#### Solution
This PR updates the ModelListPresenter class in ModelListPresenter.kt to remove the automatic remote repository check at startup, allowing the application to launch successfully even without network connectivity.
#### Changes
- Modified ModelListPresenter.kt to eliminate the blocking remote repository check during initialization
- Ensures the app can start and display locally available models without requiring network access
#### Benefits
- Improved offline experience: Users can now use the app with locally downloaded models even without internet connection
- Faster startup time: Eliminates unnecessary network delays during app launch
- Better error handling: Prevents indefinite loading states caused by network issues
#### Testing
- [x] App launches successfully without internet connection
- [x] Local models are displayed correctly
- [x] No indefinite loading screen when remote repository is unreachable
Fixes the startup hanging issue when remote model repository is not accessible.

### DIFF
--- a/apps/Android/MnnLlmChat/app/src/main/java/com/alibaba/mnnllm/android/modelist/ModelListPresenter.kt
+++ b/apps/Android/MnnLlmChat/app/src/main/java/com/alibaba/mnnllm/android/modelist/ModelListPresenter.kt
@@ -54,7 +54,8 @@ class ModelListPresenter(private val context: Context, private val view: ModelLi
         if (items != null) {
             onListAvailable(items, null)
         }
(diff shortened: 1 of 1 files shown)


Juude:
just not show loading dialog,fixed in https://github.com/alibaba/MNN/pull/3585

PR #3615, opened 2025-06-11
fix(ModelDownloadManager): Update download progress check method

shunlibest:
Change the download progress check from `getDownloadSizeTotal` to `getDownloadSizeSaved` to more accurately reflect the saved download data size.

**before:**
before

**fixed:**
fixed


Juude:
this has been fixed in 0.5.1.2

PR #3723, opened 2025-07-16
new feature to allow user to specify cpu

jules-ai:
I’m submitting this PR to introduce a new configuration option: `MNN::BackendConfig::cpuIds`. This allows users to explicitly specify CPU core IDs to maximize multi-threaded performance by core assignment.

This idea comes from various situations we’ve observed while using MNN in our AI development. In recent years, models have become increasingly large, multi-model collaboration (often referred to as workflows) has become more common, and devices now feature more and more CPU cores — even mobile devices are starting to offer double-digit core counts. However, as the number of threads exceeds four, the performance gains from MNN’s multi-threading start to diminish.

In version 2.9.3 (c6f25cafc63cc89996709a2ec1cb2ddca96ca5a5), MNN added a `threadNumber` parameter to its thread pool, enabling multiple instances to share the thread pool while using different thread counts. However, I believe this is a poor practice for two reasons:
1. The initialization of thread count relies on querying the number of CPU cores, which sometimes fails (e.g., on Windows), rendering this feature ineffective in many cases.
2. When two instances/sessions share a thread pool (counting on the instance with fewer threads), resource contention is inevitable.

Therefore, this PR can also be seen as a bugfix addressing these issues.

With this PR, you can expect improvements in more than following scenarios:

- Multi-process use: You can assign your UI and backend algorithms to separate CPU cores, reducing context-switching overhead.
- Multi-instance use: You can allocate cores based on each instance’s QPS demands. For example, in a shipped product, our face detection model requires 25 FPS for real-time performance, while face recognition only needs 1 FPS. Without explicit core binding, both models would compete for resources uncontrollably.
- Multi-session use: For throughput-sensitive workloads, you can now run four single-threaded instances, each pinned to a different core — providing better throughput than a single four-threaded model.
- On high-core-count devices: I have a 64-core server. Previous versions of MNN would spawn 64 threads, with most sitting idle. Once a second instance starts, it competes for the only and few active threads. This PR resolves that by giving precise control over core and thread placement.

### DIFF
--- a/docs/inference/session.md
+++ b/docs/inference/session.md
@@ -185,13 +185,10 @@ struct BackendConfig {
         void* sharedContext = nullptr;
         size_t flags; // Valid for CPU Backend
     };
-
-    /** user specified cpu cores */
-    std::vector<int> cpuIds;
 };
 ```
 
-`memory`、`power`、`precision`分别为内存、功耗和精度偏好。支持这些选项的后端会在执行时做出相应调整；若不支持，则忽略选项。`cpuIds`允许用户指定一组CPU核心用于计算，但这些指定会被严格的校验合法性，当线程数量与合法的指定核心数量相同时得到最佳性能。
+`memory`、`power`、`precision`分别为内存、功耗和精度偏好。支持这些选项的后端会在执行时做出相应调整；若不支持，则忽略选项。
 
 示例：
 后端 **OpenCL**
--- a/docs/inference/session.md
+++ b/docs/inference/session.md
@@ -181,11 +181,17 @@ struct BackendConfig {
     PrecisionMode precision = Precision_Normal;
     
     /** user defined context */
-    void* sharedContext = nullptr;
+    union {
+        void* sharedContext = nullptr;
+        size_t flags; // Valid for CPU Backend
+    };
+
+    /** user specified cpu cores */
+    std::vector<int> cpuIds;
 };
 ```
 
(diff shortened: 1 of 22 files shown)


docs/inference/session.md
```diff
@@ -181,11 +181,17 @@ struct BackendConfig {
...
     
     /** user defined context */
-    void* sharedContext = nullptr;
+    union {
+        void* sharedContext = nullptr;
+        size_t flags; // Valid for CPU Backend
+    };
+
+    /** user specified cpu cores */
+    std::vector<int> cpuIds;
```
jxt1234: 是否有办法不增加这里的成员变量。比如换用 setSessionHint 的方式设置 CPU

jules-ai: 目前setSessionHint的接口只支持int作为值参数，如果用int作为bitmap，又要考虑int在不同环境下的位数和cpu数量的判别的问题，导致逻辑更复杂了，而且用bit位来设置也不直观😵‍💫目前没有找到更好的办法

jules-ai:
已将cpu核心设定从`BackendConfig`移到`HintMode`，在x64（Ubuntu）测试 Session/Module X THREAD_POOL/OPENMP 均验证绑定成功。

jules-ai:
以下工具均增加了命令行参数cpuIds以便测试使用：
image

PR #3730, opened 2025-07-17
optimze MnnLlmChat build progress,no need to execute shell script

FranzKafkaYu:
In this PR,I changed `MnnLlmChat` build progress,which will be more friendly to Android App developers.  
- `libMNN.so` no need to be compiled via `build_64.sh`,this will be compiled in JNI compilation.  
- all of these CMake flags will be passed via gradle  during build progress.    
- modified README for this PR   

With this PR,we can build `MnnLlmChat` easily：  
- In Windows,we can use AndroidStuido build this App directly.  
- In Linux,we can use gradle cmd tool build this App directly.  
NO NEED TO RUN `build_64.sh`.


FranzKafkaYu:
@jxt1234 Hey, could you check out this PR when you get a chance，I'm pretty sure it's working fine.

jxt1234:
@Juude

Juude:
1. really  need to distinct linux and windows? ,  would it better to replace with "build from comand line and build from android studio"?
2. keep the old built method by add a variable 'BUILT_FROM_PREBUILT', so that we can build from a prebuilt lib.

FranzKafkaYu:
> 1. really  need to distinct linux and windows? ,  would it better to replace with "build from comand line and build from android studio"?
> 2. keep the old built method by add a variable 'BUILT_FROM_PREBUILT', so that we can build from a prebuilt lib.

1.In the current build process, we need to execute the build_64.sh script to obtain libMNN.so, but since it's a bash script, it can't be executed directly on Windows. Therefore, it's better to distinguish between Linux and Windows environments. For app developers, building the APK using Android Studio is more user-friendly, as they may not be familiar with cmd tools.  

2.Add  `MNN_BUILD_LLM_FROM_PREBUILT` variable to keep the old build method,default value is `OFF`      
```  
2.1 In Windows,no cmd tools,place the prebuilt `libMNN.so` in `project\android\build_64\lib`, and modify the `gradle.properties` file in the `MnnLlmChat` project to configure `MNN_BUILD_LLM_FROM_PREBUILT=ON`.

2.2 In Linux with cmd tools ,follow these instrcutions:   
  cd project/android
  mkdir build_64
  cd build_64
  ../build_64.sh "-DMNN_LOW_MEMORY=true -DMNN_CPU_WEIGHT_DEQUANT_GEMM=true -DMNN_BUILD_LLM=true -DMNN_SUPPORT_TRANSFORMER_FUSE=true -DMNN_ARM82=true -DMNN_USE_LOGCAT=true -DMNN_OPENCL=true -DLLM_SUPPORT_VISION=true -DMNN_BUILD_OPENCV=true -DMNN_IMGCODECS=true -DLLM_SUPPORT_AUDIO=true -DMNN_BUILD_AUDIO=true -DMNN_BUILD_DIFFUSION=ON -DMNN_SEP_BUILD=OFF -DCMAKE_INSTALL_PREFIX=."
  make install
  cd ../../../apps/Android/MnnLlmChat
  ./gradlew assembleDebug -PMNN_BUILD_LLM_FROM_PREBUILT=ON
```    
Of course you  also can replace the prebuilt `libMNN.so` in  `project\android\build_64\lib`

Juude (approved):

FranzKafkaYu:
@jxt1234 Hi,this PR is ready for merge :wink:

PR #3731, opened 2025-07-17
toolbar color changed

venkadesh4321:
Able to build the project and updated the color of toolbar


Juude:
why change the background.

PR #3751, opened 2025-07-23
[Fix] missing first output token in llm.cpp generate function

secret-ai-dev:
Fixed an issue where the generate function would skip the first token during generation. The problem occurred because the function wasn't properly handling the initial token sampling phase.

### DIFF
--- a/transformers/llm/engine/src/llm.cpp
+++ b/transformers/llm/engine/src/llm.cpp
@@ -625,10 +625,12 @@ std::vector<int> Llm::generate(MNN::Express::VARP input_embeds, int max_tokens) 
     _t.reset();
     mContext->current_token = sample(logits);
     mContext->sample_us += _t.durationInUs();
+    mContext->history_tokens.push_back(mContext->current_token);
+    mContext->output_tokens.push_back(mContext->current_token);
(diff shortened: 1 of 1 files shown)


jxt1234:
What's the case for missing initial token? The first token will be push in Generaters.

secret-ai-dev:
> What's the case for missing initial token? The first token will be push in Generaters.

In our use case with version a739ea5, this calling pattern results in missing the first token:

```cpp
Llm::generate_init();
Llm::generate(inputs, 1);

// At this point, context->current_token is not the first output token, but the second token
```

The issue occurs because after `generate_init()`, the first call to `generate()` with `max_new_tokens = 1` should return the initial sampled token, but instead it advances to the next token before returning. This means we lose the first token that was actually generated.

jxt1234:
OK. I get it. But the pr may cause bug for MTP and LookAhead. Let me find better way to fix.

jxt1234:
See if https://github.com/alibaba/MNN/pull/3753/ can solve the bug?

secret-ai-dev:
> See if #3753 can solve the bug?

Yes! Thank you!

PR #4117, opened 2026-01-17
[BugFix] Fix quantized.out default batch size for fixed-shape models

LudovicoYIN:
### Motivation
The `quantized.out` tool defaults to batch size 32 when `batch_size` is not set.
Some models (e.g., YOLO26n) use fixed-shape Reshape constants with batch=1, which
causes a Reshape mismatch and can lead to a crash during quantization.

### Changes
Track whether `batch_size` is explicitly set in the config
If not set and the model input batch is fixed, use the model batch instead of 32

### Testing
Reproduced the failure on YOLO26n with default batch
Confirmed quantization completes after using model batch by default

Fixes: [MNN 3.3.0 量化 YOLO26 模型时发生段错误 (Segmentation Fault)  https://github.com/alibaba/MNN/issues/4116](https://github.com/alibaba/MNN/issues/4116)

### DIFF
--- a/tools/quantization/calibration.cpp
+++ b/tools/quantization/calibration.cpp
@@ -475,6 +475,7 @@ Calibration::Calibration(MNN::NetT* model, const uint8_t* modelBuffer, const int
     }
     if (picObj.HasMember("batch_size")) {
         _batch = picObj["batch_size"].GetInt();
+        _batchSetByUser = true;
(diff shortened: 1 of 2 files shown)


LudovicoYIN:
Thanks for the review and for merging this! Glad to contribute.

PR #4133, opened 2026-01-28
fix: interleaved mrope in qwen3-vl

hebangwen:
This PR aims to address #4120 #4104 . Qwen3-VL uses [interleaved mrope](https://github.com/huggingface/transformers/blob/v4.57.6/src/transformers/models/qwen3_vl/modeling_qwen3_vl.py#L299-L314) instead of [blocked mrope](https://github.com/huggingface/transformers/blob/v4.57.6/src/transformers/models/qwen2_5_vl/modeling_qwen2_5_vl.py#L577-L583) in Qwen2.5VL. The old impl behaves the same when processing text, but causes infinite loop in vision mode.


wangzhaode:
Thanks for the PR. We referenced your solution to fix this issue while adding support for Qwen3.5.

Issue #4212, opened 2026-03-05
Qwen3-Embedding-0.6B generate_llm_qnn.py 报错


kk928290341:
成功导出mnn模型

导出qnn时报错
python npu/generate_llm_qnn.py --model /p2b_share/MNN-master/Qwen3-Embedding-0.6B-MNN/ --soc_id=52 --dsp_arch=v73
Step1: Make IO
Segmentation fault (core dumped)

Cost:  0.4871361255645752  s
Step2: Seperate Model
model: /p2b_share/MNN-master/model/llm.mnn
Segmentation fault (core dumped)
Cost:  0.39046382904052734  s
Step3: Compile to QNN
/p2b_share/qnn-r23600.1
Traceback (most recent call last):
  File "/p2b_share/MNN-master/transformers/llm/export/../../../build/../source/backend/qnn/npu_convert.py", line 10, in <module>
    with open(sys.argv[1]) as f:
         ~~~~^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: 'npu_postreat.json'
Cost:  0.83382248878479  s
Step4: Move result file to  /p2b_share/MNN-master/model/
End

EdisonPeee:
Target Device: QCS8550

和你遇到一樣的問題, 使用 Llama-3.1-TAIDE-LX-8B-Chat 成功導出 mnn 模型, 接續導 QNN 出錯

```
python3 generate_llm_qnn.py \
    --model /.../.../.../.../MNN_models/mnn_models/Llama-3.1-TAIDE-LX-8B-Chat \
    --soc_id 43 \
    --dsp_arch v73 \
    --mnn_path /.../.../.../MNN/build \
    --chunk_size 128
Step1: Make IO
Segmentation fault (core dumped)

Cost:  23.44584059715271  s
Step2: Seperate Model
model: /.../.../.../.../MNN_models/mnn_models/Llama-3.1-TAIDE-LX-8B-Chat/llm.mnn
Segmentation fault (core dumped)
Cost:  0.18132662773132324  s
Step3: Compile to QNN
/opt/qcom/qairt/2.42.0.251225
Traceback (most recent call last):
  File "/.../.../.../MNN/build/../source/backend/qnn/npu_convert.py", line 10, in <module>
    with open(sys.argv[1]) as f:
FileNotFoundError: [Errno 2] No such file or directory: 'npu_postreat.json'
Cost:  0.06715011596679688  s
[...]

PR #4216, opened 2026-03-06
[LLM:Bugfix] Fix Qwen3-Embedding QNN export crash

LudovicoYIN:
### Motivation
`generate_llm_qnn.py` assumes all exported models use generative LLM IO.
This breaks embedding models such as `Qwen3-Embedding-0.6B`, which use
`sentence_embeddings` as output and do not have `logits_index`.

As a result:
- Step1 crashes in `generateLlmIO`
- Step2 runs with missing test IO
- Step3 reports a misleading `npu_postreat.json` not found error

### Changes
Detect embedding models from `llm_config.json`
Generate embedding test IO in `generateLlmIO`
Keep the existing generative LLM path unchanged
Add missing test IO checks in `compilefornpu`
Stop the QNN export script immediately when a step fails

### Testing
Reproduced the failure on `Qwen3-Embedding-0.6B`
Confirmed Step1 generates `testdir/1` and `testdir/128` successfully
Confirmed Step2 generates `npu_postreat.json` successfully
Confirmed Step3 proceeds into QNN graph compilation instead of failing at the original crash point

Fixes: [Qwen3-Embedding-0.6B generate_llm_qnn.py 报错 #4212](https://github.com/alibaba/MNN/issues/4212)

### DIFF
--- a/tools/cpp/compilefornpu.cpp
+++ b/tools/cpp/compilefornpu.cpp
@@ -900,12 +900,20 @@ int main(int argc, const char* argv[]) {
             for (auto iter = testdir.Begin(); iter != testdir.End(); iter++) {
                 std::string dirname = iter->GetString();
                 auto subinputs = MNN::Express::Variable::load((dirname + "/input.mnn").c_str());
+                if (subinputs.empty()) {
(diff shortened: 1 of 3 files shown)


wangzhaode:
@jxt1234

PR #4248, opened 2026-03-12
Add header for uint8_t and fix build with gcc 15.2.

ioctl-user:
Fix MNN cli build with the gcc 15.2 without this patch compiler gives error, because uint8_t is undefined.


wangzhaode:
tokenizer 代码做了重构，请检查一下新版本是否还存在这个问题

PR #4323, opened 2026-03-26
[OpenCL:Feature] Accelerate MmapAllocator::read by replacing per-call lseek+read with mmap-based memcpy

Riverspring996:
## Changes
The cached file is already mapped at onAlloc time, so each read can be a plain memcpy from the pre-established mapping rather than issuing lseek+read syscalls, eliminating both the kernel syscall and the kernel-to-user copy.
## Validation
Measured total wall-clock time of OpenCLMmapAllocator::read during model loading via Android ATrace, with multiple warm-up launches to exclude cold page-fault. Results confirm a reduction about 30% in OpenCLMmapAllocator::read and 14% in total loading.

read_duration

memcpy_duration

- Device: vivo X100/V2309A
- Soc: Dimensity 9300
- Kernel: Linux 6.1.124


Qxinyu:
Thank you for your code commit. We have already addressed this issue internally, and the latest code has now been pulled.

PR #4348, opened 2026-04-06
[Vulkan:Bugfix] Removed static lookup to dynamic symbol

EduMenges:
## Description

Removed static lookup of dynamic symbol to prevent linker errors on debug, static builds.

`vkGetPhysicalDeviceFeatures2KHR` is a legacy symbol that is only recognized by 1.0 loaders. The current code tries to dynamically set it by verifying if it is null, which doesn't work in static linked contexts because functions are always defined. This leads to a problem that only occurs in debug builds, because it tries to statically link to a non-existent symbol. This doesn't happen on release mode because once the compiler sees that `if (!getFeatures2)` will never be false, it shaves the second lookup of the code and the linker doesn't even attempt to look for it. What should have been done is to look for it dynamically instead of statically.

## Module

Vulkan

## Type

- [ ] Feature
- [X] Bugfix
- [ ] Perf
- [ ] Refact
- [ ] Style
- [ ] Doc
- [ ] Test
- [ ] Chore

## Checklist

### DIFF
--- a/source/backend/vulkan/component/VulkanDevice.cpp
+++ b/source/backend/vulkan/component/VulkanDevice.cpp
@@ -209,15 +209,6 @@ VulkanDevice::VulkanDevice(std::shared_ptr<VulkanInstance> instance)
     vkEnumerateDeviceExtensionProperties(mPhysicalDevice, nullptr, &extensionCount, nullptr);
     std::vector<VkExtensionProperties> extensions(extensionCount);
     vkEnumerateDeviceExtensionProperties(mPhysicalDevice, nullptr, &extensionCount, extensions.data());
-
-    VkPhysicalDeviceShaderFloat16Int8FeaturesKHR float16Features = {};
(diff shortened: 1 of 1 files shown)


WhiSKy22 (commented):
Thanks for the fix. The vkGetPhysicalDeviceFeatures2 / vkGetPhysicalDeviceFeatures2KHR part looks valid to me and addresses the static-link debug build issue in a reasonable way.

I would suggest narrowing the PR to only keep the functional changes:

- keep the removal of the dead code in VulkanDevice::VulkanDevice

- keep the changes in VulkanDevice::checkFP16 and VulkanDevice::checkCoopMat that remove direct symbol references to vkGetPhysicalDeviceFeatures2 / vkGetPhysicalDeviceFeatures2KHR and replace
them with dynamic lookup

Please drop the whitespace / indentation / formatting-only changes that have no functional effect. They make the diff larger than necessary and increase review burden.

WhiSKy22 (approved):

PR #4450, opened 2026-05-15
Fix ScheduleConfig.numThread ignored due to union copy bug in GCC 4.9

gaoshuzhendanten:
## Summary
- `ScheduleConfig.numThread` setting is silently ignored when using `createSession()` with GCC 4.9 (NDK r14b), causing unexpected multi-threaded inference even when `numThread=1`

## Root Cause
`numThread` is defined in an anonymous union with a default member initializer (`= 4`). When `createSession()` calls `createMultiPathSession({config})`, the `{config}` initializer_list triggers copy construction where GCC 4.9 incorrectly resets the union member to its default value (4), discarding the user-set value.

## Fix
Replace `{config}` initializer_list construction with explicit vector construction followed by reassignment of `numThread` from the original config reference.

## Reproduction
1. Set `config.numThread = 1` and call `net->createSession(config)`
2. On a 2-core ARM Android device (NDK r14b / GCC 4.9), observe:
   - CPU usage ~95% (both cores saturated) instead of expected ~50%
   - `strace -f -e trace=clone` shows a worker thread created via `clone(CLONE_THREAD)`
   - `/proc/<pid>/task/` shows 2 threads, both in R state with `sched_yield` spin-wait from ThreadPool
3. Debug prints confirm: `config.numThread=1` before `createSession`, but `config.numThread=4` inside `createMultiPathSession` after the vector copy

## Test Plan
- Verified on ARM Android device (2-core, NDK r14b / GCC 4.9):
  - Before fix: 2 threads, CPU 93-98%, ThreadPool active
  - After fix: 1 thread, CPU 50%, no clone syscall, no ThreadPool initialized

### DIFF
--- a/source/core/Interpreter.cpp
+++ b/source/core/Interpreter.cpp
@@ -351,11 +351,15 @@ Session* Interpreter::createMultiPathSession(const std::vector<ScheduleConfig>& 
 }
 
 Session* Interpreter::createSession(const ScheduleConfig& config) {
-    return createMultiPathSession({config});
+    std::vector<ScheduleConfig> configs(1, config);
+    configs[0].numThread = config.numThread;
+    return createMultiPathSession(configs);
 }
 
 Session* Interpreter::createSession(const ScheduleConfig& config, const RuntimeInfo& runtime) {
-    return createMultiPathSession({config}, runtime);
+    std::vector<ScheduleConfig> configs(1, config);
+    configs[0].numThread = config.numThread;
+    return createMultiPathSession(configs, runtime);
 }
 
 bool Interpreter::releaseSession(Session* session) {


gaoshuzhendanten:
## Minimal Reproduction Program

```cpp
#include <cstdio>
#include <vector>
#include <string>

struct Config {
    std::vector<std::string> saveTensors;
    int type = 0;
    union {
        int numThread = 4;
        int mode;
    };
    struct Path {
        std::vector<std::string> inputs;
        std::vector<std::string> outputs;
        int mode = 0;
    };
    Path path;
    int backup = 0;
    void* backendConfig = nullptr;
};

int main() {
    Config c;
    c.numThread = 1;
    printf("original: c.numThread = %d\n", c.numThread);

    Config c2(c);
    printf("copy ctor: c2.numThread = %d (expect 1)\n", c2.numThread);

    std::vector<Config> vec;
    vec.push_back(c);
    printf("push_back: vec[0].numThread = %d (expect 1)\n", vec[0].numThread);

    std::vector<Config> vec2 = {c};
    printf("init_list: vec2[0].numThread = %d (expect 1)\n", vec2[0].numThread);

    auto fn = [](std::vector<Config> v) { return v[0].numThread; };
    int result = fn({c});
    printf("fn({c}):   result = %d (expect 1)\n", result);

    int bugs = (c2.numThread != 1) + (vec[0].numThread != 1) +
               (vec2[0].numThread != 1) + (result != 1);
    if (bugs == 0)
        printf("\nPASS: all copies preserve numThread correctly\n");
    else
        printf("\nFAIL: %d copy path(s) reset numThread to default\n", bugs);
    return bugs;
}
```

## GCC 4.9 (NDK r14b, arm-linux-androideabi-g++ 4.9) Output

```
original: c.numThread = 1
copy ctor: c2.numThread = 4 (expect 1)
push_back: vec[0].numThread = 4 (expect 1)
init_list: vec2[0].numThread = 4 (expect 1)
fn({c}):   result = 4 (expect 1)

FAIL: 4 copy path(s) reset numThread to default
```

## GCC 11.4 (Ubuntu 11.4.0-1ubuntu1~22.04.3, x86_64) Output

```
original: c.numThread = 1
copy ctor: c2.numThread = 1 (expect 1)
push_back: vec[0].numThread = 1 (expect 1)
init_list: vec2[0].numThread = 1 (expect 1)
fn({c}):   result = 1 (expect 1)

PASS: all copies preserve numThread correctly
```

## Analysis

The bug triggers when the struct is **not trivially copyable** (contains members like `std::vector<std::string>` that require a non-trivial copy constructor). In this case, GCC 4.9 generates an implicit copy constructor that incorrectly applies the NSDMI (Non-Static Data Member Initializer) to the anonymous union member during copy construction — resetting it to the default value (`4`) instead of copying from the source object.

When the struct is trivially copyable (POD members only), GCC 4.9 uses memcpy for the copy and the bug does **not** trigger.

The C++ standard is clear that NSDMIs should only be used during default construction, not during copy construction. The implicit copy constructor must perform member-wise copy. GCC 4.9 violates this rule; GCC 5+ handles it correctly.

Reference: [In C++0x, do non-static data member initializers override the implicit copy constructor?](https://stackoverflow.com/questions/8218397/in-c0x-do-non-static-data-member-initializers-override-the-implicit-copy-cons)

wangzhaode:
Thanks for submitting this PR and the detailed analysis!

After evaluation, we believe this fix is not necessary to merge into the main branch for the following reasons:

1. **GCC 4.9 / NDK r14b is extremely outdated** — Google completely removed GCC from the NDK starting with r18 (2018) and switched entirely to Clang. MNN no longer targets GCC 4.9 as a supported compiler.

2. **This is essentially a compiler bug workaround** — The C++ standard clearly specifies that the implicit copy constructor should perform member-wise copy. This is a known defect in GCC 4.9's handling of anonymous unions with NSDMIs, and is not present in GCC 5+ or Clang.

If your project specifically requires using MNN with NDK r14b / GCC 4.9, we recommend maintaining this patch in your local repository. Thank you again for your contribution!

PR #4476, opened 2026-05-27
Fix CUDA Softmax echo artifacts and MatMul precisionType bug

cgisky1980:
### Problem

The `SOFTMAX_AXIS_REDUCE` kernel uses `blockReduceMax`/`blockReduceSum` for block-level reduction. When the softmax axis is small (e.g., 34 in our Stable Audio 3 Decoder), many threads in a block are idle but still participate in the reduction. These idle threads read uninitialized `__shared__` memory values, causing subtle perturbations in softmax weights.

In attention layers, these perturbations cause cross-position information leakage, producing audible echo artifacts — the model speaks normally but with a gradually fading echo/noise overlay.

### Reproduction

We discovered this bug while running Stable Audio 3 inference with MNN CUDA backend:
1. Convert the Stable Audio 3 Decoder (SAME-S architecture) to MNN format
2. Run inference on CUDA backend
3. The output audio has echo artifacts at positions corresponding to attention softmax with axis=34
4. CPU backend produces clean audio

### Fix

For `axis <= 64`, use the simple `SOFTMAX` kernel where each thread independently computes a complete softmax, avoiding block reduction entirely:

```cpp
if (axis <= 64) {
    SOFTMAX<<<block_num, threads_num>>>((const float*)input, (float*)dst, inside, axis, outside, count);
    checkKernelErrors;
} else if (static_cast<CUDABackend*>(backend())->useFp16()) {
    // existing SOFTMAX_AXIS_REDUCE path
}
```

Also added explicit template instantiations for `SOFTMAX`, `SOFTMAX_WARP_32`, and `SOFTMAX_AXIS_REDUCE` (both float and half) to ensure other translation units (e.g., `AttentionExecution.cu`) can still link.

### Future Improvement

A more robust fix would be to add a thread-active mask to `blockReduceMax`/`blockReduceSum` so that only threads with valid indices participate in the reduction. This would fix `SOFTMAX_AXIS_REDUCE` for all axis sizes without needing the fallback. The current threshold of 64 is conservative — the simple kernel works well for small axes, while `SOFTMAX_AXIS_REDUCE` remains efficient for large axes.

---

### Problem

In `MatMulExecution.cu` line 589, `mInfo.precisionType` is hardcoded to `2` (FP16_FP16), even when running in FP16FP32Mix or FP32 mode. This forces the MatMul to always compute in pure FP16, causing precision anomalies in models that expect mixed FP16/FP32 computation.

### Fix

```cpp
// Before:
mInfo.precisionType = 2;  // Always FP16_FP16

// After:
mInfo.precisionType = (mFp16Fp32MixInfer || mFp32Infer) ? 0 : 2;  // FP16_FP32 when mixed, FP16_FP16 otherwise
```

This ensures that when `Precision_Normal` (FP16FP32Mix) or `Precision_High` (FP32) is selected, the MatMul uses the appropriate precision type instead of always falling back to pure FP16.

---

## Testing

Both fixes have been validated in production with Stable Audio 3 inference:
- **Softmax fix**: 120-second audio generation with no echo artifacts, RTF 28.4x on RTX 2080 Ti
- **MatMul fix**: Correct FP16FP32Mix behavior verified across NC, DiT, and Decoder models
- All MNN CUDA models (NumberConditioner, DiT, Decoder) produce outputs with max_diff < 0.002 vs CPU reference

### DIFF
--- a/source/backend/cuda/execution/MNNCUDAFunction.cuh
+++ b/source/backend/cuda/execution/MNNCUDAFunction.cuh
@@ -96,7 +96,7 @@ T blockReduceMax(T val)
     }
     __syncthreads();
 
-    val = (threadIdx.x < (blockDim.x >> 5 )) ? shared[lane] : (T)0.0f;
+    val = (threadIdx.x < (blockDim.x >> 5 )) ? shared[lane] : (T)(-1.0e30f);
     val = warpReduceMax(val);
     return val;
 }
--- a/source/backend/cuda/execution/MatMulExecution.cu
+++ b/source/backend/cuda/execution/MatMulExecution.cu
@@ -586,7 +586,7 @@ void MatMulExecution::setArguments(const std::vector<Tensor *> &inputs, const st
             }
             mInfo.epilogueVectorize = true;
             mInfo.epilogueType = 0;// Linear
-            mInfo.precisionType = 2;// FP16_FP16
+            mInfo.precisionType = (mFp16Fp32MixInfer || mFp32Infer) ? 0 : 2;// FP16_FP32 or FP16_FP16
             mInfo.backend = mBackend;
 
             if(mUseRRLayout) {
--- a/source/backend/cuda/execution/SoftmaxExecution.cu
+++ b/source/backend/cuda/execution/SoftmaxExecution.cu
@@ -183,7 +183,6 @@ ErrorCode SoftmaxExecution::onResize(const std::vector<Tensor *> &inputs, const 
     mCpuParam.outside = outside;
     mCpuParam.axis = input->length(axis);
 
-    // printf("\nsoftmax:%d-%d-%d, %d-%d\n", mCpuParam.inside, mCpuParam.outside, mCpuParam.axis, mNeedUnpackC4, axis);
     return NO_ERROR;
 }
 
@@ -198,9 +197,6 @@ ErrorCode SoftmaxExecution::onExecute(const std::vector<Tensor *> &inputs, const
         dst = (void*)mStorage.deviceId();
     }
 
-    //MNN_PRINT("softmax input dims:%d, size:%d-%d-%d-%d\n", inputs[0]->dimensions(), inputs[0]->batch(), inputs[0]->height(), inputs[0]->width(), inputs[0]->channel());
-    //MNN_PRINT("softmax storage dims:%d, size:%d-%d-%d-%d\n", mStorage.dimensions(), mStorage.batch(), mStorage.height(), mStorage.width(), mStorage.channel());
-
     auto runtime = static_cast<CUDABackend*>(backend())->getCUDARuntime();
     int inside = mCpuParam.inside;
     int outside = mCpuParam.outside;
@@ -208,7 +204,11 @@ ErrorCode SoftmaxExecution::onExecute(const std::vector<Tensor *> &inputs, const
     int count = inside * outside;
     int block_num = runtime->blocks_num(count);
     int threads_num = runtime->threads_num();
-    if (static_cast<CUDABackend*>(backend())->useFp16()) {
+
+    if (axis <= 64) {
+        SOFTMAX<<<block_num, threads_num>>>((const float*)input, (float*)dst, inside, axis, outside, count);
+        checkKernelErrors;
+    } else if (static_cast<CUDABackend*>(backend())->useFp16()) {
 	if(axis % 256 == 0 || axis >= 768) {
             block_num = count;
             int calc_multi_num = (axis + 255) / 256;
@@ -249,6 +249,7 @@ ErrorCode SoftmaxExecution::onExecute(const std::vector<Tensor *> &inputs, const
             checkKernelErrors;
         }
     }
+
     if (mNeedUnpackC4) {
         backend()->onCopyBuffer(&mStorage, outputs[0]);
     }
@@ -271,5 +272,12 @@ public:
 };
 
 static CUDACreatorRegister<SoftmaxCreator> __init(OpType_Softmax);
+
+template __global__ void SOFTMAX<float>(const float*, float*, const int, const int, const int, const int);
+template __global__ void SOFTMAX_WARP_32<float>(const float*, float*, const int, const int, const int, const int);
+template __global__ void SOFTMAX_AXIS_REDUCE<float>(const float*, float*, const int, const int, const int, const int, const int, const int);
+template __global__ void SOFTMAX<half>(const half*, half*, const int, const int, const int, const int);
+template __global__ void SOFTMAX_WARP_32<half>(const half*, half*, const int, const int, const int, const int);
+template __global__ void SOFTMAX_AXIS_REDUCE<half>(const half*, half*, const int, const int, const int, const int, const int, const int);
 }
 }
--- a/source/backend/cuda/execution/SoftmaxExecution.cu
+++ b/source/backend/cuda/execution/SoftmaxExecution.cu
@@ -207,14 +207,7 @@ ErrorCode SoftmaxExecution::onExecute(const std::vector<Tensor *> &inputs, const
 
     bool isHalf = (inputs[0]->getType().bits == 16);
 
-    if (axis <= 64) {
-        if (isHalf) {
-            SOFTMAX<<<block_num, threads_num>>>((const half*)input, (half*)dst, inside, axis, outside, count);
-        } else {
-            SOFTMAX<<<block_num, threads_num>>>((const float*)input, (float*)dst, inside, axis, outside, count);
-        }
-        checkKernelErrors;
-    } else if (isHalf) {
+    if (isHalf) {
         if(axis % 256 == 0 || axis >= 768) {
             block_num = count;
             int calc_multi_num = (axis + 255) / 256;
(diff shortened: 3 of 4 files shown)


wangzhaode (commented):
**AI 评审员建议** (本评论由 AI 辅助审查代码生成，供参考)

---

**Review: Fix CUDA Softmax echo artifacts and MatMul precisionType bug**

Two well-documented fixes. Analysis:

### Bug 1: Softmax (SoftmaxExecution.cu)

**Root cause**: `SOFTMAX_AXIS_REDUCE` uses block-level reduction where idle threads read uninitialized shared memory when `axis` is small.

**Fix**: For `axis <= 64`, use the simple `SOFTMAX` kernel that does per-thread independent computation.

Observations:
- The threshold of 64 is conservative and well-justified
- The explicit template instantiations at the bottom ensure linkage from other TUs — correct approach

**Issue found**: The new `if (axis <= 64)` branch is placed **before** the `useFp16()` check, and always casts to `(const float*)input, (float*)dst`. When `useFp16()` is true and `axis <= 64` (e.g., the axis=34 case mentioned in the PR description), this passes half-precision device memory to a float kernel, which will produce incorrect results.

The fix should be:
```cpp
if (axis <= 64) {
    if (static_cast<CUDABackend*>(backend())->useFp16()) {
        SOFTMAX<<<block_num, threads_num>>>((const half*)input, (half*)dst, inside, axis, outside, count);
    } else {
        SOFTMAX<<<block_num, threads_num>>>((const float*)input, (float*)dst, inside, axis, outside, count);
    }
    checkKernelErrors;
} else if (static_cast<CUDABackend*>(backend())->useFp16()) {
    ...
```

**Evidence**: Current master (line 211) shows the `useFp16()` branch uses `(const half*)` / `(half*)` casts. The PR's new branch at the top always uses `(const float*)` / `(float*)` regardless of precision mode.

### Bug 2: MatMul precisionType (MatMulExecution.cu:589)

**Fix**: `mInfo.precisionType = (mFp16Fp32MixInfer || mFp32Infer) ? 0 : 2;`

Verified in master:
- `mFp16Fp32MixInfer` set at line 297: `(precisonLevel == 0)`
- `mFp32Infer` set at line 296: `(precisonLevel == 1)`
- Current code hardcodes `precisionType = 2` at line 589

Logic is correct.

### Removed debug prints

Cleanup of commented-out `printf`/`MNN_PRINT` — good.

### Summary

The MatMul fix is correct. The Softmax fix has a **type-safety bug** in the `axis <= 64` branch that needs to be addressed before merge. The author's described test case (Stable Audio 3 with axis=34) likely works only because that model runs in FP32 mode; under FP16 this would produce garbage output.

**Recommendation**: request changes (Softmax FP16 path missing).

wangzhaode (changes requested):
**AI 评审员建议** (本评论由 AI 辅助审查代码生成，供参考)

---

**Update: FP16 type-safety bug confirmed with runtime test**

I built this PR on a Linux CUDA 12.9 + RTX 3090 (sm_86) machine and wrote a targeted test for the `axis <= 64` branch under FP16 precision mode.

**Test**: Softmax with shape=[4, 34], axis_dim=34, `Precision_Low` (FP16)

**Result**:
```
  max_diff=45119.984375 mean_diff=nan
  CPU row0 sum=1.000000, CUDA row0 sum=nan
  CPU[0:5] = 0.0026 0.0085 0.0359 0.0123 0.0052
  CUDA[0:5]= -0.0000 1.2266 -0.0006 1.2266 -2.9648
  RESULT: FAIL
```

**Root cause**: The new `if (axis <= 64)` branch (line ~207 in the PR diff) always calls:
```cpp
SOFTMAX<<<...>>>((const float*)input, (float*)dst, ...)
```
This executes **before** the `useFp16()` check, so when FP16 mode is active, the kernel interprets `half` memory as `float`, producing garbage.

**Fix needed**: Add FP16 dispatch inside the `axis <= 64` branch:
```cpp
if (axis <= 64) {
    if (static_cast<CUDABackend*>(backend())->useFp16()) {
        SOFTMAX<<<block_num, threads_num>>>((const half*)input, (half*)dst, inside, axis, outside, count);
    } else {
        SOFTMAX<<<block_num, threads_num>>>((const float*)input, (float*)dst, inside, axis, outside, count);
    }
    checkKernelErrors;
} else if (...) {
```

**Note**: The MatMul precisionType fix (Bug 2) in this PR is correct and verified. Only the Softmax FP16 path needs fixing.

**Environment**: Ubuntu 24.04, CUDA 12.9, RTX 3090 (sm_86), built with `-DMNN_CUDA=ON -DMNN_CUDA_NATIVE_ARCH=ON`

wangzhaode:
## Test Report: FP16 Softmax Bug Confirmed

**Environment**: Ubuntu 24.04, CUDA 12.9, RTX 3090 (sm_86), built with `-DMNN_CUDA=ON -DMNN_CUDA_NATIVE_ARCH=ON -DMNN_SUPPORT_TRANSFORMER_FUSE=ON`

### Test case

Custom C++ test: `_Softmax` on shape `[4, 34]` with `Precision_Low` (FP16 mode), triggering the new `axis <= 64` branch.

### Result

```
Softmax FP16 test: shape=[4,34], axis_dim=34 (<=64 branch)
  max_diff=45119.984375 mean_diff=nan
  CPU row0 sum=1.000000, CUDA row0 sum=nan
  CPU[0:5] = 0.0026 0.0085 0.0359 0.0123 0.0052
  CUDA[0:5]= -0.0000 1.2266 -0.0006 1.2266 -2.9648
  RESULT: FAIL
```

### Root cause

The new `if (axis <= 64)` branch unconditionally casts to `(const float*)` and `(float*)`. When `Precision_Low` (FP16) is active, the actual device memory contains `half` data, so the `SOFTMAX` kernel reads garbage.

### Suggested fix

```cpp
if (axis <= 64) {
    if (static_cast<CUDABackend*>(backend())->useFp16()) {
        SOFTMAX<<<block_num, threads_num>>>((const half*)input, (half*)dst, inside, axis, outside, count);
    } else {
        SOFTMAX<<<block_num, threads_num>>>((const float*)input, (float*)dst, inside, axis, outside, count);
    }
    checkKernelErrors;
} else if (static_cast<CUDABackend*>(backend())->useFp16()) {
    // ... existing path
```

### Note

The MatMul `precisionType` fix (Bug 2) in this PR is correct — no issues found there. Only the Softmax FP16 path needs to be fixed before merge.

cgisky1980:
@wangzhaode 

Fixed the FP16 type-safety issue. Instead of useFp16(), we use inputs[0]->getType().bits == 16 for more precise tensor type detection — useFp16() checks backend config which may not reflect actual tensor data type. This also includes the blockReduceMax inactive thread default value fix (0.0f → -1.0e30f).

cgisky1980:
Thanks for the thorough review and the runtime test @wangzhaode — the FP16 type-safety bug was real and the test results made it crystal clear.

I've updated the PR with a refactored fix. Key changes:

### 1. Removed `axis <= 64` workaround

The `axis <= 64` branch was a conservative workaround, not a root-cause fix. The `SOFTMAX_AXIS_REDUCE` kernel already initializes idle threads with correct identity values (`-FLT_MAX` for max, `0.0f` for sum), so the kernel is correct for all axis sizes once the underlying reduction functions are fixed. I removed the workaround entirely — the code now follows the same dispatch logic for all axis sizes.

### 2. Type detection: `getType().bits` instead of `useFp16()`

```cpp
bool isHalf = (inputs[0]->getType().bits == 16);
```

`useFp16()` checks the backend's configured precision mode, which doesn't always reflect the actual tensor data type. `getType().bits` directly inspects the tensor, so it's correct regardless of how the backend was configured. This was already documented as a fix in our fork.

### 3. `blockReduceMax` identity value

```cpp
// MNNCUDAFunction.cuh, line 99
val = (threadIdx.x < (blockDim.x >> 5)) ? shared[lane] : (T)(-1.0e30f);
// Before: ... : (T)(0.0f);
```

In `blockReduceMax`, threads beyond the valid warp count participate in the final warp-level max reduction. For a max operation, the identity element is `-∞`, not `0`. When all real max values happen to be negative (which occurs in attention softmax with certain data distributions), an idle thread contributing `0.0f` incorrectly wins the reduction, corrupting the softmax normalization. This is the actual root cause that the `axis <= 64` workaround was papering over.

### PR now consists of 3 commits

1. `1203a48` — Original: MatMul `precisionType` fix (unchanged, verified correct)
2. `3fa25ad` — Softmax: `getType().bits` type detection + `blockReduceMax` identity fix
3. `d6e2eb1` — Softmax: remove `axis <= 64` workaround (no longer needed)

The fixes are complementary: `blockReduceMax` identity ensures the reduction is mathematically correct for all axis sizes and data distributions, and `getType().bits` ensures the correct kernel template is selected regardless of backend configuration.

wangzhaode:
Thanks for the great contribution! I've verified this PR on our RTX 3090 server (CUDA 12.9) — both the Softmax and MatMul fixes compile cleanly and pass all related tests with no regressions. The blockReduceMax identity fix and the precisionType dynamic selection are solid improvements. LGTM, ready to merge once the commit message format is fixed.

PR #4477, opened 2026-05-27
Fix CUDA backend CMake build on Windows

cgisky1980:
## Problem

Building MNN with CUDA on Windows (MSVC + CUDA 12.8) fails due to several CMake issues:

1. **C++17 not enabled by default**: The `--std=c++17` flag is only added when `MNN_SUPPORT_TRANSFORMER_FUSE` or `MNN_LOW_MEMORY` is set. However, cutlass headers (used by FMHA and other CUDA kernels) require C++17 regardless. This causes compilation errors on Windows.

2. **MSVC `__cplusplus` macro is wrong**: MSVC reports `__cplusplus` as `199711L` even when compiling with `/std:c++17`, unless `/Zc:__cplusplus` is passed. This causes cutlass `is_unsigned_v` and other C++17 feature detection to fail.

3. **Missing cuBLAS link**: On Windows, `MNN_CUDA_LIBS` only includes `${CUDA_LIBRARIES}` but not cuBLAS, causing linker errors for operations that depend on cuBLAS.

4. **Missing SM arch macro**: FMHA kernels use `MNN_CUDA_ENABLE_SM##` macros for architecture-specific code paths, but these macros are never defined in the CMake configuration.

## Fix

All changes are in `source/backend/cuda/CMakeLists.txt`:

- Always add `--std=c++17` to NVCC flags
- Add `-Xcompiler /Zc:__cplusplus` for MSVC
- Add `MNN_CUDA_ENABLE_SM${CUDA_ARCH_FLAGS_readable_code}` definition when `MNN_CUDA_NATIVE_ARCH` is enabled
- Find and link cuBLAS library on Windows

## Testing

Built successfully with CUDA 12.8 + MSVC on Windows 11. Verified with Stable Audio 3 inference (all CUDA models: NC, DiT, Decoder work correctly).


wangzhaode (commented):
**AI 评审员建议** (本评论由 AI 辅助审查代码生成，供参考)

---

**Review: Fix CUDA backend CMake build on Windows**

The PR addresses 4 related Windows CUDA build issues. Code analysis:

1. **Always enable C++17** (line 11 change): Reasonable — cutlass headers require C++17 regardless of `MNN_SUPPORT_TRANSFORMER_FUSE`/`MNN_LOW_MEMORY`. This makes the build more robust.

2. **`/Zc:__cplusplus` for MSVC** (line 19): Standard MSVC workaround for correct `__cplusplus` macro reporting. No downside.

3. **`MNN_CUDA_ENABLE_SM` macro** (lines 124-126): Defined only when `MNN_CUDA_NATIVE_ARCH` is set, which is opt-in. Safe scope.

4. **cuBLAS link on Windows** (lines 186-190): Uses `find_library` with fallback — if cuBLAS is not found, behavior is unchanged. Good defensive pattern.

Questions:
- For item 3: is `CUDA_ARCH_FLAGS_readable_code` always a single value? If targeting multiple architectures, could this produce an unexpected macro name?
- For item 1: does removing the conditional affect minimum CUDA toolkit version requirements? (C++17 in nvcc requires CUDA >= 11.0)

Overall: well-documented PR, clear problem/fix mapping. Unable to build-verify locally (no Windows + CUDA environment), but logic looks sound.

wangzhaode (approved):
**Verified on Linux (Ubuntu 24.04, CUDA 12.9, RTX 3090 sm_86)**

Build test with `-DMNN_CUDA=ON -DMNN_CUDA_NATIVE_ARCH=ON -DMNN_SUPPORT_TRANSFORMER_FUSE=ON`:

- **PR branch**: `MNN_Cuda_Main` builds successfully (100%, zero errors)
- **master**: fails with `fmha_flash_attention.h:283: #error This file can only be included one of sm 75, 80, 86 or 89 are defined.`

Root cause confirmed: master never defines `MNN_CUDA_ENABLE_SM##` when `MNN_CUDA_NATIVE_ARCH=ON`, causing FMHA kernel compilation to fail. The PR correctly adds `add_definitions(-DMNN_CUDA_ENABLE_SM${CUDA_ARCH_FLAGS_readable_code})`.

Other changes also verified via code review:
1. Unconditional `--std=c++17`: safe, CUDA 12.9 supports it; MNN already requires CUDA >= 11.0
2. `/Zc:__cplusplus` for MSVC: Windows-only, guarded by `if(WIN32)`
3. cuBLAS link on Windows: uses `find_library` with fallback, no-op on Linux

LGTM.

wangzhaode (approved):

PR #4502, opened 2026-06-02
[Core:Bugfix] Per-RTM KVCACHE_INFO meta to fix Qwen3-VL repeating output

wangzhaode:
#4441 reset Runtime::pMeta to nullptr from the processor RTM to isolate the vision/audio encoder from LLM kvcache state, but RuntimeManager::setHintPtr wrote to the ExecutorScope's pooled Runtime, which is shared across all RTMs of the same backend type — so the LLM RTM's pMeta got nullified too. Subsequent CPUAttention onClone captured nullptr, KV cache stopped growing during decode, and the LLM emitted "好的好的好的..." until max_tokens.

Move pMeta ownership to RuntimeAttr (per-RTM) and push it onto the Runtime at Backend create/forward boundaries via applyMetaToRuntime(). Drop the nullptr resets in omni.cpp; isolation is now architectural.

## Type

- [ ] Feature
- [x] Bugfix
- [ ] Perf
- [ ] Refact
- [ ] Style
- [ ] Doc
- [ ] Test
- [ ] Chore

## Checklist

### DIFF
--- a/express/Executor.cpp
+++ b/express/Executor.cpp
@@ -247,10 +247,14 @@ void Executor::RuntimeManager::setExternalPath(std::string path, int type) {
     mInside->mContent->modes.setExternalPath(path, type);
 }
 void Executor::RuntimeManager::setHintPtr(Interpreter::HintMode mode, void* value) {
-    auto current = ExecutorScope::Current();
-    auto rt = current->getRuntime();
-    for (auto& iter : rt.first) {
-        iter.second->pMeta = value;
+    if (mode == Interpreter::KVCACHE_INFO) {
+        mInside->mMeta = value;
+    }
+}
+
+void Executor::RuntimeManager::applyMetaToRuntime() const {
+    for (auto& iter : mInside->mRuntime.first) {
+        iter.second->pMeta = mInside->mMeta;
     }
 }
 
--- a/express/RuntimeAttr.hpp
+++ b/express/RuntimeAttr.hpp
@@ -23,6 +23,8 @@ struct RuntimeAttr {
     RuntimeInfo mRuntime;
     std::shared_ptr<Runtime> mInfo;
     std::shared_ptr<Cache> mCache;
+    // Per-RTM KVCache meta; pushed to the (pooled) Runtime by applyMetaToRuntime().
+    void* mMeta = nullptr;
     // Use for static module to compute flops
     float mFlops;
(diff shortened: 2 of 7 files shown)


wangzhaode (commented):
## Review Summary

**Build**: ✅ Clang 编译无错误无新警告  
**Tests**: ✅ run_test.out 全部通过 (369 passed, 0 failed)

### PR Intent

Root cause 分析准确: 原 `#4441` 修复写入 `ExecutorScope::Current()->getRuntime()` 获取的 pooled Runtime,同 backend 类型的所有 RTM 共享该 Runtime,导致 LLM RTM 的 `pMeta` 被 processor RTM 的 `nullptr` reset 覆盖,KV cache 停止增长,Qwen3-VL 产生重复输出。

### Findings

**1. [Minor] Thread-safety gap in clone path**

`express/module/StaticModule.cpp:683-684`

```cpp
ctx->pRuntimeManager->applyMetaToRuntime();
```

If a cloned module shares the same `RuntimeManager` with the original and both run on different threads concurrently, `mInside->mRuntime` may be mutated while the other thread is using it.

- **Severity**: Low — not a regression; `StaticModule::clone` was not thread-safe before this PR either.
- **Suggested fix**: Consider adding a mutex inside `RuntimeManager::applyMetaToRuntime()`, or document that callers must ensure serialization when sharing a RuntimeManager across threads.
- **Action**: Optional; not a blocker.

**2. [Info] `setHintPtr` no longer writes to Runtime directly**

`express/Executor.cpp:249-253`

The previous behavior of writing `pMeta` into the Runtime immediately in `setHintPtr` is replaced by storing to `RuntimeAttr::mPMeta` first and deferring to `applyMetaToRuntime()`. This is correct but worth confirming that no other code path still expects `setHintPtr` to have immediate effect on the Runtime.

- **Status**: Verified — all callers (`Interpreter::setHintPtr` → `Executor::RuntimeManager::setHintPtr`) go through this path.
- **Action**: None.

### Verdict

The fix is well-targeted, the per-RTM `RuntimeAttr::mPMeta` design isolates the KVCACHE_INFO meta correctly, and the removal of `nullptr` resets in `omni.cpp` is a clean consequence. The test suite passes. No blockers.

**LGTM** (optional finding #1 notwithstanding).

PR #4531, opened 2026-06-12
[CPU:Bugfix] fix rvv pack and unpack functions errors

ihb2032:
## Description

fix rvv pack and unpack functions errors

## Module

CPU

## Type

- [ ] Feature
- [X] Bugfix
- [ ] Perf
- [ ] Refact
- [ ] Style
- [ ] Doc
- [ ] Test
- [ ] Chore

## Checklist

### DIFF
--- a/source/backend/cpu/riscv/rvv/MNNPackC2.cpp
+++ b/source/backend/cpu/riscv/rvv/MNNPackC2.cpp
@@ -1,74 +1,38 @@
 #include <riscv_vector.h>
 
-void MNNPackC2(float *dst, const float *src, size_t area, size_t depth, int *areaOffset) {
-    int depthC2 = depth / 2;
-    int depthRemain = depthC2 * 2;
-    int remain = depth - depthRemain;
-    const float *srcOffset = src;
-    const float *srcChannel[2];
+void MNNPackC2(float* dst, const float* src, size_t area, size_t depth, int* areaOffset) {
+    const size_t srcAreaStride = (size_t)areaOffset[0];
+    const size_t dstAreaStride = (size_t)areaOffset[1];
+    const ptrdiff_t dstStrideBytes = 2 * (ptrdiff_t)sizeof(float);
+    const size_t depthC2 = (depth + 1) / 2;
 
-    for (int z = 0; z < depthC2; ++z) {
-        float *dstZ = dst + z * areaOffset[1] * 2;
+    for (size_t z = 0; z < depthC2; ++z) {
+        const size_t cBase = z * 2;
+        float* dstZ = dst + z * dstAreaStride * 2;
+        size_t valid = depth - cBase;
+        if (valid > 2)
+            valid = 2;
 
-        for (int y = 0; y < 2; ++y) {
-            srcChannel[y] = srcOffset + areaOffset[0] * y;
-        }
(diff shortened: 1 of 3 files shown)


wangzhaode (commented):
Thanks for the cleanup @ihb2032. I reviewed the three functions against the reference scalar implementations in `source/backend/cpu/compute/CommonOptFunction.cpp` (`MNNPackC4Common` / `MNNUnpackC4Common` / `MNNPackC2Common`).

**`MNNPackC4` and `MNNUnpackC4`: LGTM.** Verified by replaying the strided load/store indexing against the reference for `depth = 4, 5, 7, 8` — outputs match exactly, including the odd-channel zero-padding tail.

**`MNNPackC2`: this looks incorrect.** In `MNNPackC2.cpp`, the real channel data is stored at offset `+y`:

```cpp
for (size_t y = 0; y < valid; ++y) {
    ...
    __riscv_vsse32_v_f32m8(dstZ + 2 * x + y, dstStrideBytes, v, vl);
}
if (valid > 0) {
    ...
    __riscv_vsse32_v_f32m8(dstZ + 2 * x + 1, dstStrideBytes, zero, vl); // overwrites channel y==1
}
```

When `valid == 2` (the common full-pair case), channel `y == 1` is written at offset `+1` and then the `if (valid > 0)` block unconditionally overwrites offset `+1` with zeros, so the second channel is lost.

Comparing against `MNNPackC2Common` for `depth = 2`, `area = 3` (no padding):

```
expected: [1, 4, 2, 5, 3, 6]
actual:   [1, 0, 2, 0, 3, 0]   // every even channel zeroed out
```

The zero-fill should only cover the missing channel (odd `depth`), mirroring your `MNNPackC4` tail:

```cpp
for (size_t y = valid; y < 2; ++y) {
    size_t x = 0;
    while (x < area) {
        const size_t vl = __riscv_vsetvl_e32m8(area - x);
        vfloat32m8_t zero = __riscv_vfmv_v_f_f32m8(0.0f, vl);
        __riscv_vsse32_v_f32m8(dstZ + 2 * x + y, dstStrideBytes, zero, vl);
        x += vl;
    }
}
```

Could you update `MNNPackC2` accordingly? Note: I verified this via index-level logic comparison against the reference implementation, not on real RVV hardware, so a check on an actual RISC-V/QEMU target would be good to confirm.

wangzhaode (commented):
Thanks for the fix @ihb2032. The updated `MNNPackC2` now zero-fills only the missing channel via `for (size_t y = valid; y < 2; ++y)`, mirroring the `MNNPackC4` tail. I re-checked it against `MNNPackC2Common` across `depth = 1..5` and `area = 1/3/8` and the outputs match (e.g. `depth=2, area=3` now produces `[1, 4, 2, 5, 3, 6]` instead of the previous all-even-zeroed result). `MNNPackC4` / `MNNUnpackC4` are unchanged and were already verified. This LGTM from a logic standpoint.

(As before, verified via index-level comparison against the reference implementation in `source/backend/cpu/compute/CommonOptFunction.cpp`, not on real RVV hardware — a quick run on a RISC-V/QEMU target would be a good final confirmation.)

wangzhaode (approved):

PR #4540, opened 2026-06-15
[Converter:Bugfix] Fix TFLite binary activation conversion

devin-lai:
## Summary
- Preserve TFLite fused activations for MUL, SUB, and DIV binary conversions.
- Lower RELU_N1_TO_1 with the correct [-1, 1] clamp and fix GREATER decomposition.
- Add a focused converter regression test for fused activations and missing options tables.

## Root cause
BinaryTflite read activation data through AddOptions, so MUL/SUB/DIV activations were dropped. The binary extra transform also mapped GREATER to GREATER_EQUAL.

## Test
- git diff --check upstream/master...HEAD
- cmake --build build_converter_test_ninja --target TestTfliteBinaryActivation -j$(sysctl -n hw.ncpu)
- ./build_converter_test_ninja/TestTfliteBinaryActivation
- cmake --build build_converter_test_static --target TestTfliteBinaryActivation -j$(sysctl -n hw.ncpu)
- ./build_converter_test_static/TestTfliteBinaryActivation

### DIFF
--- a/tools/converter/source/optimizer/tflitextra/BinaryActivation.cpp
+++ b/tools/converter/source/optimizer/tflitextra/BinaryActivation.cpp
@@ -25,8 +25,8 @@ public:
         auto activationType = static_cast<tflite::ActivationFunctionType>(extra->attr()->Get(1)->i());
         auto inputs = expr->inputs();
         MNN_ASSERT(inputs.size() == 2);
-        auto input0    = inputs[0];
-        auto input1    = inputs[1];
+        auto input0 = inputs[0];
+        auto input1 = inputs[1];
         VARP newOutput;
         switch (opType) {
             case tflite::BuiltinOperator_POW: {
@@ -74,14 +74,14 @@ public:
                 break;
             }
             case tflite::BuiltinOperator_GREATER: {
-                newOutput = _GreaterEqual(input0, input1);
+                newOutput = _Greater(input0, input1);
                 break;
             }
             case tflite::BuiltinOperator_EQUAL: {
                 newOutput = _Equal(input0, input1);
                 break;
             }
-            case tflite::BuiltinOperator_NOT_EQUAL:{
+            case tflite::BuiltinOperator_NOT_EQUAL: {
                 newOutput = _NotEqual(input0, input1);
                 break;
             }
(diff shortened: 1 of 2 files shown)


tools/converter/source/TestTfliteBinaryActivation.cpp
wangzhaode, line 1: Please delete this test file.

tools/converter/CMakeLists.txt
```diff
@@ -96,4 +96,33 @@ IF(MNN_BUILD_CONVERTER)
     target_link_libraries(TestPassManager MNNConvertDeps)
     target_link_libraries(MNNConvert MNNConvertDeps)
   ENDIF()
+
+  IF(MNN_BUILD_TEST)
```
wangzhaode: please remove `TestTfliteBinaryActivation` binary, undo the CMake change

wangzhaode (commented):
Following up on the two inline comments. To give some context on why removing the test:

- `tools/converter/source/` doesn't currently host per-op converter unit tests of this style, so adding one breaks the existing convention here.
- The new binary is built only on explicit invocation; it is not wired into `ctest` or `run_test.out`, so it would not actually guard regressions in CI as-is.
- The test couples tightly to internal converter APIs (`liteOpConverterSuit::search`, `TFliteExtraManager::find`), which adds maintenance cost without a corresponding CI signal.

The three source-level fixes themselves look correct — I verified each against current `master` and ran the full regression on arm64 macOS: `./run_test.out` → `{"passed":376,"failed":0,"blocked":0,"skipped":0}`. Once `TestTfliteBinaryActivation.cpp` and the CMake additions are removed, this is good to merge.

If you want similar coverage longer term, integrating an equivalent end-to-end check into the existing `test/` op-test framework would be a better fit.

devin-lai:
> Following up on the two inline comments. To give some context on why removing the test:
> 
> * `tools/converter/source/` doesn't currently host per-op converter unit tests of this style, so adding one breaks the existing convention here.
> * The new binary is built only on explicit invocation; it is not wired into `ctest` or `run_test.out`, so it would not actually guard regressions in CI as-is.
> * The test couples tightly to internal converter APIs (`liteOpConverterSuit::search`, `TFliteExtraManager::find`), which adds maintenance cost without a corresponding CI signal.
> 
> The three source-level fixes themselves look correct — I verified each against current `master` and ran the full regression on arm64 macOS: `./run_test.out` → `{"passed":376,"failed":0,"blocked":0,"skipped":0}`. Once `TestTfliteBinaryActivation.cpp` and the CMake additions are removed, this is good to merge.
> 
> If you want similar coverage longer term, integrating an equivalent end-to-end check into the existing `test/` op-test framework would be a better fit.

Done, thanks for the  review..

wangzhaode (approved):

PR #4547, opened 2026-06-16
[CPU:Perf] Optimize QuantizedAdd with RVV

jxgxxx:
## Description

This PR introduces pure RISC-V Vector (RVV) architectural optimizations for the CPU backend, specifically targeting the `Deconvolution` (weight transformation `MNNDeconvTransformWeightC4`) and `QuantizedAdd` operators. 

**Performance Benchmarks:**
* **Hardware:** SG2044 64-Core RISC-V 64-bit
* **OS:** PolyOS Server 24.03 LTS

| Function / Operator | Input Dimensions & Configuration | Baseline (C++) | RVV Optimized | Speedup |
| :--- | :--- | ---: | ---: | ---: |
| `CPUQuantizedAdd` | `N=1, H=112, W=112, C=64, elements=802816, 1-thread, iter=200` | 15.3202 ms | 3.55873 ms | **4.30x** |

## Module

CPU (specifically RVV backend)

## Type

- [ ] Feature
- [ ] Bugfix
- [x] Perf
- [ ] Refact
- [ ] Style
- [ ] Doc
- [ ] Test
- [ ] Chore

## Checklist

### DIFF
--- a/source/backend/cpu/CPUDeconvolution.cpp
+++ b/source/backend/cpu/CPUDeconvolution.cpp
@@ -21,6 +21,10 @@
 //#define MNN_OPEN_TIME_TRACE
 #include <MNN/AutoTime.hpp>
 
+#if defined(__riscv) && defined(MNN_USE_RVV)
+extern void MNNDeconvTransformWeightC4_RVV(const float* src, float* dst, int outputCount, int srcCount, int area);
+#endif
+
 namespace MNN {
 
 CPUDeconvolutionBasic::CPUDeconvolutionBasic(int inputChannel, const Op* convOp, Backend* b)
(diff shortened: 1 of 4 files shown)


wangzhaode:
Thanks for the RVV optimizations. The Deconvolution weight-layout logic looks good, but two issues need to be addressed before merging.

1. I tested the PR on a riscv64 board with GCC 15.2, `MNN_USE_RVV=ON`, and `MNN_SUPPORT_DEPRECATED_OP=ON`. The build fails in `CPUQuantizedAdd.cpp` at `functions()->supportRVV` because `CoreFunctions` is only forward-declared. Please include `compute/CommonOptFunction.h` or otherwise make the complete definition available.
2. Please make the RVV fixed-point rounding exactly match the existing scalar/NEON implementation. Using `vsmul`/`vssra` with RNU differs for some negative tie cases. An integer-formula comparison over 500,000 randomized cases found two uint8 outputs differing by 1 LSB. Please add coverage for negative intermediate values and rounding boundaries.

The Deconvolution C4/tail layout simulation passed. It would also be helpful to clarify the multi-thread benchmark methodology because the reported per-iteration numbers are not directly consistent with the single-thread results. Thanks!

jxgxxx:
Thanks for the detailed review. I updated the PR in commit `0bb67203` and rebased it onto the current `master` (`c425199d`).

Changes addressing the review:

1. Added `compute/CommonOptFunction.h`, so `CoreFunctions::supportRVV` is accessed through the complete type and the native riscv64 build succeeds with `MNN_USE_RVV=ON` and `MNN_SUPPORT_DEPRECATED_OP=ON`.
2. Removed the `vssra` RNU shortcut and implemented `RoundingDivideByPOT` with the same remainder/threshold formula as the scalar/NEON path, including negative ties. `vsmul` RNU is retained for `SaturatingRoundingDoublingHighMul`, whose signed rounding and saturation semantics match the existing helper. The new test covers focused negative/tie boundaries and 516,736 randomized elements across 32 quantization configurations with exact byte-for-byte comparison on SG2044.
3. Removed the previous synthetic multi-thread benchmark because it created/joined workers per iteration and did not model MNN's persistent thread pool. The updated PR description reports only the reproducible single-core benchmark: 15.3202 ms to 3.55873 ms per call (`4.30x`) for `N=1, H=112, W=112, C=64`.

I also re-evaluated the Deconvolution change against the actual existing RVV `MNNPackC4` / `MNNPackCUnit` path rather than a hand-written scalar reference. It measured only `1.061x` for OC=128 and `0.999x` for OC=130, so the specialized Deconvolution path and its benchmark claims have been removed from this PR.

The final diff is limited to `CPUQuantizedAdd.cpp`, the RVV kernel. Please take another look when convenient.

PR #4550, opened 2026-06-16
[Infra:Refact] Modernize shader build scripts to handle complex paths

RinZ27:
## Description

Refactored several `makeshader.py` scripts across Vulkan, Metal, and OpenGL backends to replace legacy shell execution patterns. The previous implementation relied on `os.popen` with string concatenation, which I found could fail when encountering file paths with spaces or special characters. 

By switching to `os.walk` for file discovery and using `subprocess.run` with list arguments, these scripts are now significantly more robust across different environments. This change also cleans up the temporary file management during the SPIR-V assembly process. I have verified the changes locally by generating shaders for each affected backend, and the resulting C++ source files remain consistent with the original logic.

## Module

Infra, Vulkan, Metal, OpenCL

## Type

- [ ] Feature
- [ ] Bugfix
- [ ] Perf
- [x] Refact
- [ ] Style
- [ ] Doc
- [ ] Test
- [ ] Chore

## Checklist

### DIFF
--- a/source/backend/metal/makeshader.py
+++ b/source/backend/metal/makeshader.py
@@ -2,15 +2,17 @@
 import sys
 import os
 import re
+import subprocess
+
 gOutputHeadFile = "AllShader.hpp"
 gOutputSourceFile = "AllShader.cpp"
+
 def findAllShader(path):
-    cmd = "find " + path + " -name \"*.metal\""
-    vexs = os.popen(cmd).read().split('\n')
     output = []
-    for f in vexs:
-        if len(f)>1:
-            output.append(f)
+    for root, dirs, files in os.walk(path):
+        for file in files:
+            if file.endswith(".metal"):
+                output.append(os.path.join(root, file))
     return output
 
 def getName(fileName):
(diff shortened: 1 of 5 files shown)


wangzhaode:
I verified this locally by running the modified scripts against the original ones across all affected backends:

| Backend | File discovery | Generated output (MD5) |
|---------|---------------|----------------------|
| Metal | 18 files, set equal ✅ | AllShader.hpp + .cpp identical ✅ |
| OpenGL | 34 files, set equal ✅ | AllShader.hpp + .cpp identical ✅ |
| Vulkan Image | 49 files, set equal ✅ | — |
| Vulkan Render | 18 files, set equal ✅ | — |

Also confirmed the space-in-path fix works as intended: old `os.popen("find " + path)` fails completely (0 files found) when the path contains spaces, while the new `os.walk` approach handles it correctly.

Traversal order is consistent with `find` on the current flat shader directory structures.

LGTM from a functional standpoint.

wangzhaode (approved):

PR #4557, opened 2026-06-22
[Converter:Refact] Allow disabling converters

DavidVentura:
## Description

Allows disabling specific converters for MNNConvert.
Need to migrate users of legacy ONNX models to MNN, without having them redownload everything

## Module

Converter

## Type

- [ ] Feature
- [ ] Bugfix
- [ ] Perf
- [x] Refact
- [ ] Style
- [ ] Doc
- [ ] Test
- [ ] Chore

## Checklist


wangzhaode:
Hi @DavidVentura, thanks for the contribution!

MNNConvert is an offline tool that doesn't ship to end-user devices, so binary size reduction isn't a concern for it. We'd prefer to keep the converter build system simple and don't plan to merge this.

If you have a specific need to slim down the converter binary, we'd suggest maintaining a local patch to strip out the converters you don't need.

Thanks again for the effort!

PR #4566, opened 2026-06-24
[LLM:Bugfix] Fix WikiText calibration dataset loading

LudovicoYIN:
## Description

Fix LLM calibration dataset loading with recent `datasets` / `huggingface_hub` versions.

Recent HuggingFace Hub clients require dataset repo IDs to include a namespace. The old `load_dataset("wikitext", ...)` path now fails with an invalid HF URI error. This updates built-in WikiText calibration loading to `Salesforce/wikitext`.

Also updates `download_data.py` to accept dataset specs like `Salesforce/wikitext/wikitext-2-raw-v1` by treating the final path component as the dataset config and the preceding components as the repo ID.

## Module

LLM

## Type

- [ ] Feature
- [x] Bugfix
- [ ] Perf
- [ ] Refact
- [ ] Style
- [ ] Doc
- [ ] Test
- [ ] Chore

## Checklist

### DIFF
--- a/transformers/llm/collect/README.md
+++ b/transformers/llm/collect/README.md
@@ -35,7 +35,7 @@ python get_max_values.py -m <mnn_model_path> [options]
 
 **Arguments:**
 - `-m, --mnn-path`: Path to MNN model config (required)
-- `-d, --eval_dataset`: Dataset for evaluation (default: 'wikitext/wikitext-2-raw-v1')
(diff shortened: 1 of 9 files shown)


wangzhaode (commented):
Thanks for the fix\! The root cause is confirmed — recent HuggingFace Hub versions require a namespaced repo ID, and the old `load_dataset('wikitext', ...)` call fails with `HfUriError`. The updated `Salesforce/wikitext` identifier and the new three-part spec parsing are both correct.

LudovicoYIN:
Thanks your review.

PR #4575, opened 2026-06-28
Vulkan:Speed: Speed up vulkan for corp-matrix, complete segment ops

jxt1234:
segment        Vulkan buffer    2194.10 tok/s    86.32 tok/s
non-segment    Vulkan buffer    1100 tok/s    55 tok/s

## Type

- [ ] Feature
- [ ] Bugfix
- [ ] Perf
- [ ] Refact
- [ ] Style
- [ ] Doc
- [ ] Test
- [ ] Chore

## Checklist


wangzhaode:
Thanks for the substantial Vulkan buffer optimization. The reported gains (prefill 1100 → 2194.10 tok/s, decode 55 → 86.32 tok/s) look promising.

Could you please make a few updates before we merge?

1. Please rebase onto the latest master and resolve the current conflicts.
2. Please keep the shader-generation changes focused. The current diff changes subprocess.run back to os.popen in three makeshader.py files and also includes changes under skills/. Please revert these unless they are required for this optimization, or briefly explain why they are needed.
3. Please add the basic benchmark context (device/GPU, model, quantization, and build flags) together with a simple CPU vs Vulkan correctness result.
4. Please confirm that changing the MNN_VULKAN_IMAGE default to OFF and unconditionally enabling MNN_GPU_RECORD_BATCH are intentional. The embedded SPIR-V size grows from about 1.41 MB to 1.94 MB (about +527 KB), so a brief binary-size note would also help.

No exhaustive device matrix is required; one representative real-device correctness/performance result is enough. Thanks!

PR #4579, opened 2026-06-29
[ARM:Bugfix] Enable MNNRankOneUpdate fallback on Apple Silicon

lanling-47:
## Description

The `MNNRankOneUpdateDefault` C++ fallback was guarded by `#ifndef __aarch64__`, which excludes Apple Silicon (which also defines `__aarch64__`). However, the NEON assembly path for `MNNRankOneUpdate` is not wired up for this op on Apple platforms, causing a missing implementation at link/runtime.

Fix by changing the guard to `#if !defined(__aarch64__) || defined(__APPLE__)`, so the fallback is compiled on Apple Silicon as well.

**Tested on**: Apple M4 (ARM64, Apple Silicon)

## Module

- [x] ARM

## Type

- [x] Bugfix

## Checklist

- [x] Commit message format: `[Module:Type] Description`
- [x] Compiles without errors
- [x] Tested on Apple M4
- [x] No unrelated style changes

### DIFF
--- a/source/backend/cpu/compute/CommonOptFunction.cpp
+++ b/source/backend/cpu/compute/CommonOptFunction.cpp
@@ -3913,7 +3913,7 @@ void MNNVectorTop1Int32(int32_t* input, int32_t* maxValue, int32_t* maxIndex, si
 
 #endif
 
-#ifndef __aarch64__
+#if !defined(__aarch64__) || defined(__APPLE__)
 static void MNNRankOneUpdateDefault(float* S, const float* k, const float* delta, size_t dk, size_t dv) {
     for (size_t i = 0; i < dk; ++i) {
         float k_val = k[i];


wangzhaode:
Thanks for the fix. The missing-symbol problem is real, but using __APPLE__ as the fallback condition is too broad and would disable the existing ARM64 NEON implementation for every Apple build.

The same guard covers three functions: MNNRankOneUpdateDefault, MNNDualMatVecDefault, and MNNDecayRankOneUpdateDefault. Therefore this change makes all three use the scalar C++ fallback on Apple ARM64, not only the function mentioned in the title, and can regress LinearAttention performance.

I verified on an Apple M4 that the current CMake path enables AArch64 assembly, compiles source/backend/cpu/arm/arm64/MNNRankOneUpdate.S successfully, and exports all three symbols. MNN.podspec also already includes this assembly source. The missing integration appears to be project/ios/MNN.xcodeproj, which does not reference MNNRankOneUpdate.S.

Please fix the affected build integration directly by adding MNNRankOneUpdate.S to the appropriate Apple ARM64 target, and also check whether MNNRankOneUpdateFp16.S is missing. If some Apple integration intentionally cannot compile the assembly, please add an explicit build macro indicating whether these assembly kernels are available and enable the fallback only for that configuration, rather than for all __APPLE__ builds.

Please also update the title/description to reflect all three affected functions, rebase onto current master, and provide the exact failing build command plus validation for CMake, CocoaPods, and the iOS Xcode project.

PR #4583, opened 2026-06-30
[QNN:Bugfix] Fix StridedSlice NC4HW4 layout and improve QNNPerf robustness

lanling-47:
## Summary
Fix QNN StridedSlice operator for NC4HW4 tensor format and add null-pointer guards to QNNPerf infrastructure calls.

## StridedSlice Improvements
- **NC4HW4 support**: Remap NCHW ranges to NHWC order that QNN expects when input tensor is in NC4HW4 format
- **Optional stride**: Handle `fromType=1` with 4 inputs (no stride tensor) — default stride to 1
- **Crash prevention**: Replace `MNN_ASSERT` with graceful `nullptr` returns + diagnostic `MNN_PRINT` logging
- **Null-check**: Validate `StridedSliceParam` before accessing

## QNNPerf Improvements
- Add null-pointer guards for `deviceGetInfrastructure` result
- Check `infraType` and `createPowerConfigId` before use
- Guard all `setPowerConfig` / `destroyPowerConfigId` calls against null function pointers and zero config IDs
- Prevents crashes when HTP perf infrastructure is unavailable (e.g. on some device configurations)

## Testing
Tested on Snapdragon 8 Gen 3 and 8 Elite devices with YOLO and MobileCLIP models that use StridedSlice with NC4HW4 tensors.

### DIFF
--- a/source/backend/qnn/backend/QNNPerf.cpp
+++ b/source/backend/qnn/backend/QNNPerf.cpp
@@ -16,13 +16,30 @@ QNNPerf::QNNPerf(const QNN_INTERFACE_VER_TYPE * qnnInterface) {
     mQnnInterface = qnnInterface;
 
     QnnDevice_Infrastructure_t deviceInfra = nullptr;
-    CALL_QNN(mQnnInterface->deviceGetInfrastructure(&deviceInfra));
+    auto infrastructureStatus = mQnnInterface->deviceGetInfrastructure(&deviceInfra);
+    if (infrastructureStatus != QNN_SUCCESS || deviceInfra == nullptr) {
+        MNN_PRINT("MNN_QNN: Failed to get HTP perf infrastructure, error:%lu infra=%p\n",
+                  (unsigned long)infrastructureStatus,
+                  deviceInfra);
+        return;
+    }
     QnnHtpDevice_Infrastructure_t *htpInfra  = static_cast<QnnHtpDevice_Infrastructure_t *>(deviceInfra);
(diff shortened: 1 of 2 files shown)


wangzhaode:
Thanks for the contribution. The QNNPerf guard changes are useful and the NC4HW4 StridedSlice support is needed, but there are several issues to address before merging:

1. The NC4HW4 path remaps ranges from NCHW to NHWC, but beginMask, endMask, and shrinkAxisMask are still passed in the original NCHW bit order. For example, a channel mask on bit 1 must become the last-axis bit after conversion to NHWC. Please remap every supported axis mask consistently with the ranges.
2. computeRangesType1 removes the previous length assertion without adding runtime validation. If begin/end/axes/stride lengths differ, or an axis is outside [-rank, rank), the current loop can read or write out of bounds. Please validate these inputs and reject unsupported cases cleanly.
3. The ONNX converter can produce fromType=1 StridedSlice with 3, 4, or 5 inputs. This PR supports only 4 or 5. Please support the 3-input form where both axes and steps are omitted, or explicitly explain why it must remain unsupported.
4. The unconditional parameter MNN_PRINT in the Creator will log for every StridedSlice during normal inference. Please remove it or guard it with QNN_VERBOSE.
5. Please initialize mPowerConfigId to 0 in QNNPerf.hpp so all early-return and destruction paths have deterministic state.
6. This branch is 72 commits behind master. Please rebase and preserve the unequal Slice fix from #4626, which modifies the same file.

Please add focused device validation for NC4HW4 slicing on N/C/H/W axes, begin/end/shrink masks, 3/4/5-input forms, negative axes, malformed parameter lengths, and unavailable QNN performance infrastructure. The current environment here has no connected QNN device, so these findings are based on static review.

PR #4584, opened 2026-06-30
[QNN:Feature] Auto ADSP_LIBRARY_PATH setup and device creation fallback

lanling-47:
## Summary
Robustness improvements for QNN HTP backend initialization to handle diverse device configurations gracefully.

### 1. Auto ADSP_LIBRARY_PATH Setup
- Detect `libQnnHtp.so` location at runtime via `dladdr`
- Set `ADSP_LIBRARY_PATH` to prioritize app-bundled skel libraries over system ones
- Prevents "Skel lib id mismatch" errors caused by stub/skel version conflicts between app-bundled QNN SDK and system-preinstalled libraries

### 2. Device Creation Fallback
- When default `deviceCreate` fails, exhaustively try multiple SOC/Arch/SignedPD combinations:
  - SM8750 (Snapdragon 8 Elite) / HTP v79
  - SM8650 (Snapdragon 8 Gen 3) / HTP v75
  - Both signed and unsigned process domain variants
- Continue without device handle rather than aborting entirely

### 3. Runtime Creation Safety
- Check `logHandle` and `backendHandle` are non-null before creating `QnnRuntime`
- Warn (but proceed) if `deviceHandle` is null

### 4. Minor Cleanup
- Rename `gContext.QnnInterface` → `gContext.interface` for clarity
- Use `__fp16` directly instead of `mnn_qnn_fp16_t` wrapper
- CMakeLists: fallback to bundled `3rdParty/` directory for QNN SDK headers when `QNN_SDK_ROOT` is not set

## Testing
Tested on Snapdragon 8 Gen 3 and Snapdragon 8 Elite devices. Verified that ADSP path auto-detection resolves skel mismatch issues and device fallback works when default config fails.

### DIFF
--- a/source/backend/qnn/CMakeLists.txt
+++ b/source/backend/qnn/CMakeLists.txt
@@ -4,17 +4,14 @@ file(GLOB BACKEND_SRCS ${CMAKE_CURRENT_LIST_DIR}/backend/*)
 file(GLOB EXECUTION_SRCS ${CMAKE_CURRENT_LIST_DIR}/execution/*)
 set(MNN_QNN_SRCS ${BACKEND_SRCS} ${EXECUTION_SRCS})
 
-# QNNBackend.cpp 中使用了 flatbuffers/flexbuffers，需要 flatbuffers util.cpp 中提供的
-# ClassicLocale::instance_ 静态成员定义。MSVC 链接时不会自动从其它 target 获取，因此
-# 显式将 util.cpp 一并编译进 MNN_QNN OBJECT 库。
-if (MSVC OR WIN32)
-    list(APPEND MNN_QNN_SRCS ${CMAKE_SOURCE_DIR}/3rd_party/flatbuffers/src/util.cpp)
-endif()
-
-# Prefer CMake variable QNN_SDK_ROOT (passed via -DQNN_SDK_ROOT=...), fallback to environment
+# Prefer CMake variable QNN_SDK_ROOT (passed via -DQNN_SDK_ROOT=...), fallback to environment,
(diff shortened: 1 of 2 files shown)


wangzhaode:
Thanks for the contribution. The overall direction makes sense, but there are several blocking issues to address before merging:

1. This change removes the existing Windows/MSVC compatibility path and unconditionally uses dlfcn.h, setenv, and __fp16. Please preserve the Windows implementation, including QnnHtp.dll loading and the RPC memory fallback.
2. A null deviceHandle is now treated as a valid fallback, but QnnRuntime::create still uses deviceHandle == nullptr as the initialization check. This can repeatedly recreate and overwrite the global log/backend handles. Please introduce a separate initialization state, such as std::once_flag or an explicit state flag.
3. Please do not overwrite an existing ADSP_LIBRARY_PATH or unconditionally prepend /data/local/tmp. Preserve the user-provided value, append/prepend only the detected Skel directory as needed, and check the setenv result.
4. The CMake fallback path source/backend/qnn/3rdParty does not exist in this PR or the repository, so the bundled-SDK fallback is currently ineffective. Please remove it or provide a valid dependency path.
5. Please clarify and validate the signed-PD fallback. Recent QAIRT versions default to unsigned PD and normally ship unsigned Skels, so the signed variants may not be usable without additional deployment.
6. This branch currently conflicts with master and contains a large amount of unrelated formatting/renaming. Please rebase it and keep the functional diff focused.

Please also provide validation for Windows/MSVC compilation, the normal Android deviceCreate path, each fallback path, the all-deviceCreate-failed path for both online/offline context creation, and behavior when ADSP_LIBRARY_PATH is already set.

PR #4590, opened 2026-07-02
[CPU:Bugfix] Fix dirty data reduction in RVV AccumulateSequence

ihb2032:
## Description

Fixed a bug where the final `__riscv_vfredusum` included garbage data from tail elements. Switched to `_tu` (tail undisturbed) policy in the addition loop to prevent tail clobbering during the last iteration.

## Module

CPU

## Type

- [ ] Feature
- [X] Bugfix
- [ ] Perf
- [ ] Refact
- [ ] Style
- [ ] Doc
- [ ] Test
- [ ] Chore

## Checklist

### DIFF
--- a/source/backend/cpu/riscv/rvv/MNNAccumulateSequenceNumber.cpp
+++ b/source/backend/cpu/riscv/rvv/MNNAccumulateSequenceNumber.cpp
@@ -1,17 +1,16 @@
 #include <riscv_vector.h>
+
 void MNNAccumulateSequenceNumber_RVV(float* dst, const float* src, int size) {
-    size_t vl = __riscv_vsetvlmax_e32m1();
-    vfloat32m1_t v_sum = __riscv_vfmv_v_f_f32m1(0.0f, vl);
-    int n = size;
-    for (; n > 0;) {
-        vl = __riscv_vsetvl_e32m1(n);
-        vfloat32m1_t v_src = __riscv_vle32_v_f32m1(src, vl);
-        v_sum = __riscv_vfadd_vv_f32m1(v_sum, v_src, vl);
-        n -= vl;
-        src += vl;
+    size_t vlmax = __riscv_vsetvlmax_e32m8();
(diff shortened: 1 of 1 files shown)


wangzhaode:
Thanks for fixing the RVV tail reduction issue. Using a tail-undisturbed accumulation policy is the right direction, but the current patch has two RISC-V compilation errors:

1. `__riscv_vfadd_vv_f32m8_tu` requires the destination/passthrough, both vector operands, and `vl`. This:

```cpp
acc = __riscv_vfadd_vv_f32m8_tu(acc, vs, vl);
```

should be:

```cpp
acc = __riscv_vfadd_vv_f32m8_tu(acc, acc, vs, vl);
```

2. The final reduction passes an undefined variable named `sun`:

```cpp
sum = __riscv_vfredusum_vs_f32m8_f32m1(acc, sun, vlmax);
```

It should pass `sum`:

```cpp
sum = __riscv_vfredusum_vs_f32m8_f32m1(acc, sum, vlmax);
```

The existing CI jobs do not compile this RVV source, so the green checks do not catch these errors. Please compile with a RISC-V RVV toolchain and compare against a scalar sum on the target board. It would be useful to cover `size=0`, `1`, `VLMAX-1`, `VLMAX`, `VLMAX+1`, and `2*VLMAX+3`, since those exercise both the main loop and tail handling.

The branch is also considerably behind the current `master`, so please rebase while updating the patch.

wangzhaode:
Thanks for the update. I verified the latest commit on a Spacemit K3 RISC-V board with GCC 15.2.

The RVV implementation compiled successfully, and both CPU 0 and CPU 7 passed 630 scalar-vs-RVV comparisons covering sizes 0–209, including `VLMAX-1`, `VLMAX`, `VLMAX+1`, and `2*VLMAX+3`.

The fix looks good and is ready to merge. Thanks!

wangzhaode (approved):

PR #4602, opened 2026-07-08
[CPU:Refact] Refactor the KleidiAI integration code

HenryDen:
## Description
Reshape the integration code structure to reduce legacy code.

## Type

- [ ] Feature
- [ ] Bugfix
- [ ] Perf
- [ ] Refact
- [ ] Style
- [ ] Doc
- [ ] Test
- [ ] Chore

## Checklist


wangzhaode:
Thanks for the refactor. Moving kernel selection and dispatch into the individual KleidiAI executions and removing the shared singleton is a good direction, but I think a few items need to be addressed before this can be merged:

1. Please rebase onto the latest `master`. The branch is currently conflicting and significantly behind. In particular, #4721 has now been merged and its ISA isolation must be preserved when resolving `cmake/KleidiAI.cmake`:
   - MNN integration/dispatch sources: `armv8.2-a+fp16`
   - regular KleidiAI kernels/packers: `i8mm+dotprod+fp16`
   - scalable-vector/SME2 sources: `sve+sve2`

   After this refactor, the dispatch logic lives directly in files such as `KleidiAIConvInt8.cpp` and `KleidiAIConvolution.cpp`, so those MNN integration sources must remain in the baseline ISA group. Otherwise the `SIGILL` issue fixed by #4721 may be reintroduced on Apple devices with SME2 but without SVE2.

2. The symmetric per-channel int4 path is no longer behaviorally equivalent. It changes from the dedicated `qai8dxp/qsi4cxp` kernels to the asymmetric `qsi8d32/qai4c32` kernels, and the eligibility check changes from `inputCount % 2 == 0` to `inputCount % 32 == 0`. This means shapes such as IC=16 or IC=48 that were previously accelerated will now fall back. Please clarify whether this is intentional; otherwise retain the existing path. If intentional, please include correctness and performance results for representative IC=16/32/48 cases.

3. This rewrites a large amount of packing and dispatch code without adding tests. After rebasing, please cover at least:
   - FP32 and FP16
   - symmetric/asymmetric per-channel and per-block int4
   - GEMV (`m=1`) and GEMM (`m>1`)
   - SME2 and dotprod+i8mm dispatch
   - the target iPhone no-`SIGILL` validation used for #4721

The overall cleanup direction looks reasonable, but the rebase, ISA separation, and non-equivalent int4 behavior should be resolved first.

wangzhaode:
Thanks for the update. Restoring the dedicated symmetric int4 path and adding the IC=16/32/48 cases address the earlier behavior issue.

One blocking issue remains in cmake/KleidiAI.cmake: the four MNN integration sources are now appended directly to MNN_SOURCES_KLEIDIAI, while set_source_files_properties still targets MNN_KLEIDIAI_INTEGRATION_SOURCES, which is never populated. As a result, the baseline -fno-tree-vectorize and -march=armv8.2-a+fp16 options introduced by #4721 are not applied to these dispatch sources.

Please keep KleidiAIConvolution.cpp, KleidiAIConvolutionDepthwise.cpp, KleidiAIConvInt8.cpp, and KleidiAIDenseConvolution.cpp in MNN_KLEIDIAI_INTEGRATION_SOURCES, then append that list to MNN_SOURCES_KLEIDIAI. After this fix, one smoke test on the target Apple device confirming that the KleidiAI path is actually selected and does not trigger SIGILL should be sufficient. Thanks!

wangzhaode (approved):

PR #4611, opened 2026-07-10
[LLM:Bugfix] Fix QNN export chunk-size and per-task workdir handling

LudovicoYIN:
## Description

This PR contains two small bug fixes in the LLM QNN export path.

### 1. Honor `--chunk_size` when generating prefill IO

`transformers/llm/export/npu/generate_llm_qnn.py` exposed `--chunk_size`, but `convert_llm()` always called:

```python
makeIOJson(args, 128, hidden_size, mask_type)
```

This meant the prefill QNN graph could still be generated against 128-token IO even when `--chunk_size 256/512` was requested.

### 2. Run QNN compile subprocesses inside the per-task workdir

`source/backend/qnn/npu_convert.py` invoked the compile subprocess without setting `cwd=workdir`.
For parallel conversions, this could make multiple tasks share the same process working directory assumptions and increase the chance of temporary/output path collisions.

## Why this matters

In the MNN QNN offline export path, these issues can make generated artifacts inconsistent with runtime expectations:

- prefill graph shape may not match the requested chunk size
- runtime `chunk_limits` and generated graphs may diverge
- `ppl_eval` results can become misleading
- parallel export jobs are more fragile than they need to be

## Module

LLM

## Type

- [ ] Feature
- [x] Bugfix
- [ ] Perf
- [ ] Refact
- [ ] Style
- [ ] Doc
- [ ] Test
- [ ] Chore

## Checklist

## Fixed
https://github.com/alibaba/MNN/issues/4612

### DIFF
--- a/source/backend/qnn/npu_convert.py
+++ b/source/backend/qnn/npu_convert.py
@@ -83,7 +83,7 @@ def process_src(task):
         "-t", "x86_64-linux-clang",
         "-o", workdir,
     ]
-    compile_result = run_subprocess(compile_cmd, retries=3)
+    compile_result = run_subprocess(compile_cmd, cwd=workdir, retries=3)
     if compile_result.returncode != 0:
         raise RuntimeError(f"Compile failed for src={src}: {' '.join(compile_cmd)}")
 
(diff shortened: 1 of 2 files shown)


Qxinyu (approved):

LudovicoYIN:
thanks your review.

PR #4691, opened 2026-07-30
[LLM:Bugfix] Open tokenizer file in binary mode to prevent MSVC crash

PanicTitan:
## Description

I tried running a converted model (SmolLM2-135M) using the LLM example on Windows, but it crashed right at startup with an access violation error (-1073741819) and no error message.

Then I ran the project in Visual Studio with the debugger attached to see where it was breaking. It stopped inside PipelineTokenizer::load_vocab_binary while reading tokenizer.mtok. It was trying to read a string over 55,000 characters long, which I thought it was reading random or corrupted data from the file.

In tokenizer.cpp (Tokenizer::createTokenizer), the code opens tokenizer.mtok in normal text mode first, saves the file position, and then reopens the file in binary mode using that saved position.

After searching I saw that on Windows/MSVC a position saved in text mode doesn't line up properly when used in binary mode (even if it happens to work on Linux, becasue it was working). Because of this, it jumps to the wrong byte in the file, reads garbage as the string length, and crashes.

So to fix I added `std::ios::binary` to the first `ifstream` and this fixed the offset issue and allowed the model to load and run successfully.

Note: I have tested this only on Windows 11 MSVC 2022

## Module

LLM

## Type

- [ ] Feature
- [X] Bugfix
- [ ] Perf
- [ ] Refact
- [ ] Style
- [ ] Doc
- [ ] Test
- [ ] Chore

## Checklist

- [ ] Tested on relevant platform(s) (Only windows)

### DIFF
--- a/transformers/llm/engine/src/tokenizer/tokenizer.cpp
+++ b/transformers/llm/engine/src/tokenizer/tokenizer.cpp
@@ -76,7 +76,7 @@ Tokenizer* Tokenizer::createTokenizer(const std::string& filename) {
     // AUTOTIME;
     Tokenizer* tokenizer = nullptr;
     // check file
-    std::ifstream tok_file(filename);
+    std::ifstream tok_file(filename, std::ios::binary); // std::ifstream tok_file(filename);
     if (!tok_file.good()) {
         printf("Failed: can't load tokenzier from: %s.\n", filename.c_str());
         return tokenizer;
--- a/transformers/llm/engine/src/tokenizer/tokenizer.cpp
+++ b/transformers/llm/engine/src/tokenizer/tokenizer.cpp
@@ -76,7 +76,7 @@ Tokenizer* Tokenizer::createTokenizer(const std::string& filename) {
     // AUTOTIME;
     Tokenizer* tokenizer = nullptr;
     // check file
-    std::ifstream tok_file(filename, std::ios::binary); // std::ifstream tok_file(filename);
+    std::ifstream tok_file(filename);
     if (!tok_file.good()) {
         printf("Failed: can't load tokenzier from: %s.\n", filename.c_str());
         return tokenizer;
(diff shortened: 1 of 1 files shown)


wangzhaode:
Thanks for tracking down the MSVC text/binary offset issue. The `.mtok` diagnosis and fix direction are correct, but opening the initial stream in binary mode also changes how every legacy text tokenizer is parsed on Windows.

The exporter writes `tokenizer.txt` in Python text mode, so it normally has CRLF line endings on Windows. With `std::ios::binary`, `std::getline` removes only `\n` and leaves `\r`. `HuggingfaceTokenizer::load_vocab` stores the complete line as the token and uses the complete second part of each merge rule, so the retained `\r` can corrupt vocabulary entries and merge matching.

Could you please keep the initial stream in text mode, and only reopen the file in binary mode inside the `PIPELINE` branch? The binary stream can start from the beginning, consume the magic/type line, call `load_special` on that same stream, and then call `load_vocab_binary`. This avoids reusing a text-mode `tellg()` position in a binary stream without changing legacy tokenizer behavior.

It would also be good to remove the commented-out old declaration and, if possible, test both a CRLF-header `.mtok` and a CRLF legacy HuggingFace `tokenizer.txt` on Windows. Thanks again for finding this issue!

PanicTitan:
Both test cases working on Windows/MSVC.

.mtok tested with SmolLM2-135M-MNN-full, loads and generates correctly.
Legacy tokenizer.txt tested with Hunyuan-0.5B-Instruct-MNN, loads and generates correctly.

I have "reverted" the change that I made on the tok_file openning in binary mode and now it only open this way inside the PIPELINE branch.

```cpp
Tokenizer* Tokenizer::createTokenizer(const std::string& filename) {
    // AUTOTIME;
    Tokenizer* tokenizer = nullptr;
    // check file
    std::ifstream tok_file(filename);
    if (!tok_file.good()) {
        printf("Failed: can't load tokenzier from: %s.\n", filename.c_str());
        return tokenizer;
    }
    // check tokenizer info
    std::string line;
    std::getline(tok_file, line);
    std::istringstream line_str(line);
    int magic_number, tokenizer_type;
    line_str >> magic_number;
    if (magic_number != MAGIC_NUMBER) {
        printf("Failed: magic number is wrong from: %s.\n", filename.c_str());
        return tokenizer;
    }
    line_str >> tokenizer_type;
    if (tokenizer_type == PIPELINE) {
        // .mtok binary format (PipelineTokenizer)
        tok_file.close();
        // Reopen from the beginning in binary mode instead of reusing a
        // text-mode tellg() position, that position isn't guaranteed to
        // line up with a byte offset in a separately-opened stream on
        // Windows/MSVC.
        auto* pt = new PipelineTokenizer();
        tokenizer = pt;
        std::ifstream bin_file(filename, std::ios::binary);
        std::string bin_line;
        std::getline(bin_file, bin_line);   // re-consume the magic/type line
        tokenizer->load_special(bin_file);
        pt->load_vocab_binary(bin_file);
        bin_file.close();
        tokenizer->cache_special_tokens();
        return tokenizer;
    }
    // create tokenizer
    switch (tokenizer_type)
    {
        case SENTENCEPIECE:
            tokenizer = new Sentencepiece();
            break;
        case TIKTOIKEN:
            tokenizer = new Tiktoken();
            break;
        case BERT:
            tokenizer = new BertTokenizer();
            break;
        case HUGGINGFACE:
            tokenizer = new HuggingfaceTokenizer();
            break;
        default:
            return tokenizer;
    }
    // load special tokens
    tokenizer->load_special(tok_file);
    // load vocabs
    tokenizer->load_vocab(tok_file);
    tok_file.close();
    tokenizer->cache_special_tokens();
    return tokenizer;
}
```

Let me know if there's anything else you'd like adjusted before merging.

wangzhaode (approved):

PR #4789, opened 2026-08-20
[LLM:Bugfix] Buffer split Unicode tokenizer output

primorLee:
## Description

Decode partial tokenizer output through the same tokenizer `decode()` API as the normal path. Keep at most three pending token IDs after each call, emit the oldest undecodable token when the bound is reached, and retain the suffix needed to finish a four-byte UTF-8 character.

Flush pending output before a normal decoded word and at both stop-token and maximum-length exits. Reset the buffer at the start of the next response. Literal replacement characters and malformed bytes can no longer hold later output indefinitely.

The unwired mocked tokenizer test has been removed as requested in review.

## Module
LLM

## Type
- [x] Bugfix

## Tests

- An offline probe using real Transformers 4.57.6 and tokenizers 0.22.2 passed 11 cases: split CJK, four-byte emoji, invalid/literal replacement prefixes, ASCII after pending bytes, bounded malformed input, final flushing, and the token-conversion fallback. This was a local probe; no new CI job is claimed.
- `python -m compileall -q transformers/llm/export/utils/tokenizer.py transformers/llm/export/llmexport.py`
- `python -m tabnanny transformers/llm/export/utils/tokenizer.py transformers/llm/export/llmexport.py`
- `git diff --check`

## Checklist
- [x] Commit follows the module/type convention.
- [x] Changed Python files compile.
- [x] The reported behavior was reproduced and the corrected paths were verified.

### DIFF
--- a/transformers/llm/export/llmexport.py
+++ b/transformers/llm/export/llmexport.py
@@ -177,7 +177,7 @@ class LlmExporter(torch.nn.Module):
     @torch.no_grad()
     def response(self, query):
         # self.imitate_quant()
-        self.model.decode_buffer = []
+        self.tokenizer.decode_buffer.clear()
         messages = [
             {"role": "user", "content": query}
         ]
--- a/transformers/llm/export/tests/test_tokenizer.py
+++ b/transformers/llm/export/tests/test_tokenizer.py
@@ -1,0 +1,37 @@
+from utils.tokenizer import AutoTokenizer, LlmTokenizer, PreTrainedTokenizer
+
+
+class SplitUnicodeTokenizer:
+    eos_token_id = None
+    generation_config = None
+    chat_template = None
+    vocab_size = 2
+
+    def encode(self, text, **kwargs):
+        return []
+
+    def decode(self, token_id):
+        return "\ufffd"
+
+    def convert_ids_to_tokens(self, token_ids):
+        return token_ids
+
+    def convert_tokens_to_string(self, tokens):
+        return "\u4f60" if tokens == [1, 2] else "\ufffd"
+
+    def get_vocab(self):
+        return {}
+
+
+def test_id_to_str_buffers_split_unicode_tokens(monkeypatch, tmp_path):
+    monkeypatch.setattr(PreTrainedTokenizer, "__init__", lambda self, **kwargs: None)
+    monkeypatch.setattr(
+        AutoTokenizer,
+        "from_pretrained",
+        lambda *args, **kwargs: SplitUnicodeTokenizer(),
+    )
+    tokenizer = LlmTokenizer(tmp_path, "test")
+
+    assert tokenizer.id_to_str(1) == ""
+    assert tokenizer.id_to_str(2) == "\u4f60"
+    assert tokenizer.decode_buffer == []
--- a/transformers/llm/export/utils/tokenizer.py
+++ b/transformers/llm/export/utils/tokenizer.py
@@ -19,6 +19,7 @@ class LlmTokenizer(PreTrainedTokenizer):
         self.tokenizer = tokenizer
         self.tokenizer_path = tokenizer_path
         self.model_type = model_type
+        self.decode_buffer = []
         # stop_ids
         self.stop_ids = []
         self.stop_ids.append(self.tokenizer.eos_token_id)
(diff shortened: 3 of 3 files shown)


wangzhaode:
Thanks for digging into this — the diagnosis is correct on both counts, and both are real latent bugs:

1. `self.decode_buffer` was never initialized on the tokenizer. `llmexport.py` was setting `self.model.decode_buffer` (wrong object), so the buffering branch in `id_to_str` would have raised `AttributeError` through `__getattr__` if it were ever reached. Moving the buffer onto `LlmTokenizer.__init__` is the right fix.
2. For byte-level BPE tokenizers (Qwen / GPT2 family), `tokenizer.decode()` does `bytes.decode('utf-8', errors='replace')` internally and never raises, so the whole buffering block inside `except:` was dead code and half a CJK character leaked out as `\uFFFD`. Hoisting the check out of the `try/except` is what actually makes it work.

I have one blocking concern before this can be merged, plus two smaller notes.

## Blocking: the buffer is unbounded and never flushed

I replayed the new `id_to_str` logic against a small byte-level fake vocab:

| input sequence | output | leftover buffer |
| --- | --- | --- |
| split CJK (`\xe4\xbd` + `\xa0`) | `你` | `[]` — intended fix, works |
| a token whose text really is `\uFFFD`, then `hello` | `hello` | `[4]` |
| lone continuation byte `0x80`, then `hello` twice | `hellohello` | `[5]` |
| lone continuation byte `0x80`, then split CJK | `` (empty) | `[5, 1, 2]` |

The last row is a regression. Once an ID that can never form valid UTF-8 enters the buffer (a stray continuation byte, a genuine `\uFFFD` token, `<unk>`), the buffer is never cleared, so every subsequent multi-byte character is appended and returns `''` — the rest of the response is silently swallowed. Today the same input prints garbage, which is ugly but at least visible; after this change it becomes silent data loss, which is much harder to diagnose.

Row 3 is also not quite clean: while a pending ID sits in the buffer, a following well-formed token is emitted without flushing the pending bytes, so they linger and can later concatenate with unrelated IDs.

A UTF-8 scalar is at most 4 bytes, so capping the buffer is enough:

```python
def id_to_str(self, token_id):
    token_id = int(token_id)
    try:
        word = self.tokenizer.decode(token_id)
    except Exception:
        word = self.tokenizer.convert_tokens_to_string(
            self.tokenizer._convert_id_to_token(token_id))
    if '\uFFFD' not in word:
        if self.decode_buffer:
            # pending bytes can never complete; emit them instead of dropping
            word = self._flush_decode_buffer() + word
        return word
    # Smollm-style tokenizers can emit half of a CJK character.
    self.decode_buffer.append(token_id)
    buffer_txt = self.tokenizer.decode(self.decode_buffer)
    if '\uFFFD' not in buffer_txt or len(self.decode_buffer) >= 4:
        self.decode_buffer.clear()
        return buffer_txt
    return ''
```

## Note 1: keep the retry on the same API as the fast path

The primary path uses `tokenizer.decode()`, but the buffered retry uses `convert_tokens_to_string(convert_ids_to_tokens(...))`. The latter bypasses the decoder pipeline and options such as `clean_up_tokenization_spaces`, so the two paths can disagree on the same text. Since `decode()` is what produced the partial output, `self.tokenizer.decode(self.decode_buffer)` is the consistent choice; `convert_tokens_to_string` is still a reasonable fallback when `decode()` raises.

## Note 2: the new test will not run anywhere

`transformers/llm/export/tests/` has no `__init__.py` or `conftest.py`, so `from utils.tokenizer import ...` only resolves when the working directory happens to be `transformers/llm/export`. There is also no `pytest` in `transformers/llm/export/requirements.txt` and no export test stage in `test.sh` / `test_stages.json`, so nothing in CI will execute it. Combined with the note in the PR description that the run used an import stub because `transformers` was not installed, the committed test has not actually been exercised against real `transformers`.

Also, the mock's `decode()` always returns `\uFFFD`, so the "decode succeeds" fast path and the cross-response reset that the `llmexport.py` change addresses are both uncovered.

Either wire the case into an existing CI stage (and add the `pytest` dependency), or drop the test from this PR and land it separately — the code fix is valuable on its own and I would rather not block it on test infrastructure.

## Minor

`except:` is carried over from the old code; `except Exception:` would be slightly better while you are in there. Hoisting `int(token_id)` and removing the three nested closures is a nice readability win.

Could you add the buffer cap and the pending-flush, and switch the retry to `decode()`? Happy to merge once those are in.

wangzhaode (approved):

PR #4809, opened 2026-08-26
[Converter:Bugfix] Fix divide-by-zero crashes in MNNConvert for malformed convolution/pool models

### LINKED ISSUE #4797: [bug] MNNConvert divides by zero when a model sets a convolution group, stride or dilation to 0
MNNConvert divides by zero when a model sets a convolution group, stride or dilation to 0

[attachments_mnn_divide_by_zero.zip](https://github.com/user-attachments/files/31326608/attachments_mnn_divide_by_zero.zip)

Attach: the seven reproducers listed in the table below, plus the seven sanitizer
        traces (one per reproducer)

Seven model files make `MNNConvert` die with `Floating point exception` at `19af6da`: a convolution or pooling parameter that comes straight out of the file - `group`, `strides`, `dilations`, or a channel count derived from the tensor shapes - is used as an integer divisor with no check that it is non-zero. Six distinct division sites are involved: two in the TFLite front-end and a post-conversion pass, four in the shape and CPU code that the converter's optimizer runs.

## error log

```
==216==ERROR: AddressSanitizer: FPE on unknown address 0x7fd4ca411ed5 (pc 0x7fd4ca411ed5 bp 0x7ffe34bcd480 sp 0x7ffe34bcd1c0 T0)
    #0 0x7fd4ca411ed5 in Conv2DTflite::run(...) /w/src/tools/converter/source/tflite/ConvolutionTflite.cpp:70:31
    #1 0x7fd4ca45ea5d in tflite2MNNNet(...) /w/src/tools/converter/source/tflite/liteConverter.cpp:383:22
    #2 0x7fd4c9177416 in MNN::Cli::convertModel(modelConfig&) /w/src/tools/converter/source/common/cli.cpp:644:24
    #3 0x5645ba84fb71 in main /w/src/tools/converter/source/MNNConverter.cpp:20:12

AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: FPE /w/src/tools/converter/source/tflite/ConvolutionTflite.cpp:70:31 in Conv2DTflite::run(...)
==216==ABORTING
```

The long template argument lists are shortened to `(...)` above; the untouched traces are attached. The other six are the same `FPE`, at the sites in this table. All seven pass through `MNN::Cli::convertModel`.

| reproducer | bytes | `-f` | crash site | the division |
|---|---:|---|---|---|
| `12_tflite_conv_group_divide_by_zero.bin` | 2,476 | TFLITE | `tools/converter/source/tflite/ConvolutionTflite.cpp:70` | `inputShape[3] / ci`, `ci` = `weightShape[3]` |
| `33_optimizer_group_divide_by_zero.bin` | 2,476 | TFLITE | `tools/converter/source/optimizer/postconvert/TransformGroupConvolution.cpp:191` | `srcCount % common->group` |
| `56_onnx_group_divide_by_zero.bin` | 527 | ONNX | `tools/converter/source/optimizer/postconvert/TransformGroupConvolution.cpp:191` | `srcCount % common->group` |
| `46_onnx_shape_conv_divide.bin` | 531 | ONNX | `source/shape/ShapeConvolution.cpp:86` | `/ layer->strideX()` |
| `55_onnx_shape_pool_divide.bin` | 11,844 | ONNX | `source/shape/ShapePool.cpp:74` | `/ layer->strideX()` |
| `45_onnx_conv_factory_divide.bin` | 531 | ONNX | `source/backend/cpu/compute/ConvolutionFloatFactory.cpp:269` | `common->outputCount() / group` |
| `47_onnx_conv_tiled_divide.bin` | 531 | ONNX | `source/backend/cpu/compute/ConvolutionTiledExecutor.cpp:170` | `UP_DIV(-sySta, p.dilateY)` |

## context

- MNN at `19af6da`, the converter tool `MNNConvert` (`tools/converter`)
- Built `-DMNN_BUILD_CONVERTER=ON -DCMAKE_BUILD_TYPE=Release`, clang, with AddressSanitizer on - the attached traces are that build's own output
- Linux x86-64

## how to reproduce

1. Build the converter:

```
cmake -B build -DMNN_BUILD_CONVERTER=ON -DCMAKE_BUILD_TYPE=Release
cmake --build build -j
```

2. Run it on a reproducer, with the `-f` from the table:

```
./build/MNNConvert -f TFLITE --modelFile 12_tflite_conv_group_divide_by_zero.bin \
                   --MNNModel /dev/null --bizCode check
./build/MNNConvert -f ONNX   --modelFile 46_onnx_shape_conv_divide.bin \
                   --MNNModel /dev/null --bizCode check
```

3. `Floating point exception`. The divisor is determined entirely by the file, and nothing on the path to it depends on timing or on the allocator, so the same file crashes the same way every time. One trace per reproducer is attached, from the AddressSanitizer build described above; `SIGFPE` on an integer division by zero is a CPU trap, so the sanitizer is reporting the crash rather than causing it.

## more

**Root cause.** The front-ends copy these integers into `Convolution2DCommonT` and `PoolT` exactly as the file gives them, and nothing between the copy and the division rejects zero.

`tools/converter/source/optimizer/onnxextra/OnnxConvolutionMerge.cpp`, lines 202-215:

```c++
            if (key == "dilations") {
                auto dataList = attr->list();
                dilation_h    = dataList->i()->data()[0];
                if (dataList->i()->size() >= 2) {
                    dilation_w      = dataList->i()->data()[1];
                }
            } else if (key == "group") {
                group = attr->i();
            } else if (key == "strides") {
                auto dataList = attr->list();
                stride_h      = dataList->i()->data()[0];
                if (dataList->i()->size() >= 2) {
                    stride_w      = dataList->i()->data()[1];
                }
```

and then, unconditionally, lines 267-282:

```c++
        // set param
        common->relu  = false;
        common->group = group;
        if (isDeconv) {
            common->outputCount = co * group; // deconv set inputCount to be ci, dw to be group
            common->inputCount = ci;
        } else {
            common->outputCount = co;
            common->inputCount  = ci * group; // conv set inputCount to be ci, dw to be group
        }
        common->kernelX = kw;
        common->kernelY = kh;
        common->dilateX = dilation_w;
        common->dilateY = dilation_h;
        common->strideX = stride_w;
        common->strideY = stride_h;
```

`OnnxPooling.cpp:154-160` does the same for pooling strides. The schema defaults are 1 (`schema/default/CaffeOp.fbs:14-17` and `:19`), but a default only applies when the attribute is absent - an attribute that is present and zero is stored as zero.

**The two checks that look like they cover this do not.**

`source/backend/cpu/compute/ConvolutionFloatFactory.cpp` computes the group, asserts it, and then divides by it (lines 257-269):

```c++
    int group = common->group();
    if (common->inputCount() != inputs[0]->channel() && common->inputCount() > 0) {
        group = inputs[0]->channel() / conv2d->common()->inputCount();
    }
    MNN_ASSERT(group > 0);
    if (1 == group) {
        return _createUnit(inputs[0], outputs[0], backend, op, originWeight, originWeightSize, originBias,
                           originBiasSize, quanCommon, supportSparse, lowMemory);
    }
    // TODO: Use Geometry to split
    // Split
    std::vector<std::shared_ptr<Execution>> subConvolution;
    auto groupOutputCount = common->outputCount() / group;
```

`MNN_ASSERT` expands to nothing unless `DEBUG` is defined (`include/MNN/MNNDefine.h:52`), so line 261 is not in the Release build. This one is not a zero in the file: line 259 makes `group` zero whenever the input channel count is smaller than `inputCount`.

`DCHECK` in the converter is a log statement, not a guard - `#define CHECK(x) if (!(x)) LogMessage(...).stream() << ...` (`tools/converter/include/logkit.h:69`, aliased at `:81`), with no abort and no return. The trace for `33_optimizer_group_divide_by_zero.bin` shows it: `Check failed: weightShape.size() == 4 ==> Conv2D weight ERROR!` from `ConvolutionTflite.cpp:63`, and the run carries straight on to the crash.

**Suggested fix.** Rejecting the values where they are parsed covers four of the six sites and stops a bad model reaching the shape and CPU code at all. In `OnnxConvolutionMerge.cpp`, after the attribute loop closes at line 254:

```c++
         }
+        if (group <= 0 || stride_w <= 0 || stride_h <= 0 || dilation_w <= 0 || dilation_h <= 0) {
+            MNN_ERROR("Conv: group/strides/dilations must be positive, got group %d, strides %d %d, dilations %d %d\n",
+                      group, stride_w, stride_h, dilation_w, dilation_h);
+            return nullptr;
+        }
 
         std::unique_ptr<Convolution2DT> convParam(new MNN::Convolution2DT);
```

`return nullptr` is what this function already does for an unsupported `auto_pad`, at line 225. `OnnxPooling.cpp` needs the same check on `poolParam->strideX` and `poolParam->strideY`.

The two remaining sites are not fed by a zero attribute and need their own guards. `ConvolutionFloatFactory.cpp:261`, turning the assertion that is compiled out into one that is not:

```c++
-    MNN_ASSERT(group > 0);
+    if (group <= 0) {
+        MNN_ERROR("Convolution: invalid group %d\n", group);
+        return nullptr;
+    }
```

and `ConvolutionTflite.cpp:69`, where the existing guard admits `ci == 0` instead of excluding it:

```c++
-    if (inputShape.size() == 4 && inputShape[3] > ci) {
+    if (inputShape.size() == 4 && ci > 0 && inputShape[3] > ci) {
         group = inputShape[3] / ci;
     }
```

`TransformGroupConvolution.cpp:191` is fixed by moving the `group` test above the modulo it currently follows, so that group 0 and negative groups take the path the code already takes for group 1:

```c++
         const int srcCount = common->inputCount;
+        if (common->group <= 1) {
+            iter++;
+            continue;
+        }
         const bool depthwiseLike = srcCount % common->group != 0 || common->outputCount % common->group != 0;
-        if (common->group == 1 || depthwiseLike) {
+        if (depthwiseLike) {
             iter++;
             continue;
         }
```

A zero stride is worth rejecting in `ShapeConvolution.cpp:86` and `ShapePool.cpp:74` as well - both `onComputeSize` functions already `return false` for shapes they cannot handle - so that a model arriving from any other front-end is caught too.

Found by fuzzing the model converters. Happy to send more detail if it is useful.

JingliangGao:
## Description

Fixes #4797.

`MNNConvert` crashes with a floating point exception (divide by zero) on the 7 malformed models attached to #4797, because zero/negative convolution and pooling parameters (`group`, `strides`, `dilations`) are used directly as divisors in several code paths.

This PR adds validation with clear error messages instead of crashing:

| File | Guard |
| --- | --- |
| `tools/converter/source/optimizer/onnxextra/OnnxConvolutionMerge.cpp` | reject `group < 1`, `strides < 1`, `dilations < 1` when merging ONNX convolutions (4 of the repro cases are caught here, at the frontend) |
| `tools/converter/source/optimizer/postconvert/TransformGroupConvolution.cpp` | `group <= 0` guard in both the 2D and 3D transform paths |
| `tools/converter/source/tflite/ConvolutionTflite.cpp` | `ci <= 0` guard before computing `group = inputShape[3] / ci` |
| `source/shape/ShapeConvolution.cpp` | reject non-positive `strideX/strideY/dilateX/dilateY` in shape inference |
| `source/shape/ShapePool.cpp` | reject non-positive pool `strideX/strideY` in shape inference |
| `source/backend/cpu/compute/ConvolutionFloatFactory.cpp` | replace `MNN_ASSERT(group > 0)` (a no-op in Release builds) with a real check |

## Module

Converter, Core, CPU

## Type

- [ ] Feature
- [x] Bugfix
- [ ] Perf
- [ ] Refact
- [ ] Style
- [ ] Doc
- [ ] Test
- [ ] Chore

## Checklist

### Verification

- All 7 reproduction files from #4797 now exit with a clear error instead of crashing, e.g.:
  `Conv has invalid parameters, group:0, strides:[1,1], dilations:[1,1], all must be positive`
- Valid models still convert correctly: plain / grouped / depthwise / dilated convolution, transposed convolution and pooling models all convert successfully (schema defaults for strides/dilations are 1, so there are no false positives).
- Full unit test suite (`run_test.out`) passes: **380 passed, 0 failed**.

### DIFF
--- a/source/backend/cpu/compute/ConvolutionFloatFactory.cpp
+++ b/source/backend/cpu/compute/ConvolutionFloatFactory.cpp
@@ -259,6 +259,10 @@ Execution* ConvolutionFloatFactory::create(const std::vector<Tensor*>& inputs, c
         group = inputs[0]->channel() / conv2d->common()->inputCount();
     }
     MNN_ASSERT(group > 0);
+    if (group <= 0) {
+        MNN_ERROR("Convolution %s has invalid group:%d\n", op->name()->c_str(), group);
+        return nullptr;
+    }
     if (1 == group) {
         return _createUnit(inputs[0], outputs[0], backend, op, originWeight, originWeightSize, originBias,
                            originBiasSize, quanCommon, supportSparse, lowMemory);
--- a/source/shape/ShapeConvolution.cpp
+++ b/source/shape/ShapeConvolution.cpp
@@ -27,6 +27,11 @@ public:
         MNN_ASSERT(inputs.size() >= 1);
         MNN_ASSERT(1 == outputs.size());
         const Convolution2DCommon* layer = loadCommon(op);
+        if (layer->strideX() <= 0 || layer->strideY() <= 0 || layer->dilateX() <= 0 || layer->dilateY() <= 0) {
+            MNN_ERROR("Invalid convolution parameter, strideX:%d, strideY:%d, dilateX:%d, dilateY:%d, all must be positive\n",
+                      layer->strideX(), layer->strideY(), layer->dilateX(), layer->dilateY());
+            return false;
+        }
         int kX = layer->kernelX();
         int kY = layer->kernelY();
         auto outputCount = layer->outputCount();
--- a/source/shape/ShapePool.cpp
+++ b/source/shape/ShapePool.cpp
@@ -36,6 +36,11 @@ public:
         int outw   = 1;
         int outh   = 1;
         if (!layer->isGlobal()) {
+            if (layer->strideX() <= 0 || layer->strideY() <= 0) {
+                MNN_ERROR("Invalid pool parameter, strideX:%d, strideY:%d, all must be positive\n",
+                          layer->strideX(), layer->strideY());
+                return false;
+            }
             // when given explicit pad value in tensorflow mode pool, size compute will fast failed to help find problem
             if ((layer->padType() == PoolPadType_VALID || layer->padType() == PoolPadType_SAME) && (layer->padX() != 0 || layer->padY() != 0)) {
                 MNN_PRINT("tensorflow mode pool should not have explict pad value\n");
--- a/tools/converter/source/optimizer/onnxextra/OnnxConvolutionMerge.cpp
+++ b/tools/converter/source/optimizer/onnxextra/OnnxConvolutionMerge.cpp
@@ -253,6 +253,12 @@ public:
             }
         }
 
+        if (group < 1 || stride_h < 1 || stride_w < 1 || dilation_h < 1 || dilation_w < 1) {
+            MNN_ERROR("%s has invalid parameters, group:%d, strides:[%d,%d], dilations:[%d,%d], all must be positive\n",
+                      originalOpType.c_str(), group, stride_h, stride_w, dilation_h, dilation_w);
+            return nullptr;
+        }
+
         std::unique_ptr<Convolution2DT> convParam(new MNN::Convolution2DT);
         convParam->common.reset(new MNN::Convolution2DCommonT);
         auto common = convParam->common.get();
--- a/tools/converter/source/optimizer/onnxextra/OnnxPooling.cpp
+++ b/tools/converter/source/optimizer/onnxextra/OnnxPooling.cpp
@@ -96,6 +96,11 @@ public:
             poolOp->main.type    = OpParameter_Pool;
             poolOp->main.value   = poolParam;
             poolParam->ceilModel = false;
+            // In the ONNX spec `strides` is optional and defaults to 1 per spatial axis.
+            // PoolT has no flatbuffer default for strideX/strideY (initialized to 0),
+            // so apply the spec default here explicitly.
+            poolParam->strideX = 1;
+            poolParam->strideY = 1;
             do {
                 if (type == "MaxPool") {
                     poolParam->type = MNN::PoolType_MAXPOOL;
--- a/tools/converter/source/optimizer/postconvert/TransformGroupConvolution.cpp
+++ b/tools/converter/source/optimizer/postconvert/TransformGroupConvolution.cpp
@@ -26,6 +26,10 @@ public:
             auto conv3D  = op->main.AsConvolution3D();
             auto& common = conv3D->common;
             const int srcCount = common->inputCount;
+            if (common->group <= 0) {
+                MNN_ERROR("Convolution3D %s has invalid group:%d\n", op->name.c_str(), common->group);
+                return false;
+            }
             if (common->group == 1 || op->inputIndexes.size() > 1) {
                 iter++;
                 continue;
@@ -188,6 +192,10 @@ public:
             auto conv2D  = op->main.AsConvolution2D();
             auto& common = conv2D->common;
             const int srcCount = common->inputCount;
+            if (common->group <= 0) {
+                MNN_ERROR("Convolution %s has invalid group:%d\n", op->name.c_str(), common->group);
+                return false;
+            }
             const bool depthwiseLike = srcCount % common->group != 0 || common->outputCount % common->group != 0;
             if (common->group == 1 || depthwiseLike) {
                 iter++;
--- a/tools/converter/source/tflite/ConvolutionTflite.cpp
+++ b/tools/converter/source/tflite/ConvolutionTflite.cpp
@@ -68,8 +68,10 @@ void Conv2DTflite::run(MNN::OpT* dstOp, const std::unique_ptr<tflite::OperatorT>
     const int weightSize = co * kh * kw * ci;
     if (ci <= 0) {
         MNN_ERROR("Conv2D weight has invalid input channel:%d\n", ci);
+        dstOp->type = MNN::OpType_MAX;
+        return;
     }
-    if (inputShape.size() == 4 && ci > 0 && inputShape[3] > ci) {
+    if (inputShape.size() == 4 && inputShape[3] > ci) {
         group = inputShape[3] / ci;
     }
     if (quantizedModel == 1) { // UINT8_QUANT
--- a/tools/converter/source/tflite/ConvolutionTflite.cpp
+++ b/tools/converter/source/tflite/ConvolutionTflite.cpp
@@ -66,7 +66,10 @@ void Conv2DTflite::run(MNN::OpT* dstOp, const std::unique_ptr<tflite::OperatorT>
     const int kw         = weightShape[2];
     const int ci         = weightShape[3];
     const int weightSize = co * kh * kw * ci;
-    if (inputShape.size() == 4 && inputShape[3] > ci) {
+    if (ci <= 0) {
+        MNN_ERROR("Conv2D weight has invalid input channel:%d\n", ci);
+    }
+    if (inputShape.size() == 4 && ci > 0 && inputShape[3] > ci) {
         group = inputShape[3] / ci;
     }
     if (quantizedModel == 1) { // UINT8_QUANT
--- a/tools/converter/source/torch/PoolTorch.cpp
+++ b/tools/converter/source/torch/PoolTorch.cpp
@@ -38,6 +38,10 @@ void PoolTorch::run(MNN::OpT* dstOp, const torch::jit::Node* node, TorchScope* s
                 param->strideX = 2;
                 param->strideY = 2;
             }
+        } else {
+            // PyTorch defaults stride to kernel_size when it is not specified
+            param->strideY = param->kernelY;
+            param->strideX = param->kernelX;
         }
         if (inputs.size() > 3) {
             const auto padding = getValue<std::vector<int64_t>>(inputs[3]);


JingliangGao:
@wangzhaode Hi, I have tried to solve this problem#4797. Could you please view this PR? Thanks.

wangzhaode:
Thanks a lot for this contribution, @JingliangGao — hardening the converter against divide-by-zero is definitely the right direction, and 5 of the 6 guards look correct to me. I went through each site and also did some empirical testing; there are two points I'd like to resolve before merging.

## Confirmed safe

- `source/shape/ShapeConvolution.cpp` — the guard sits before the division and covers the registered conv ops.
- `tools/converter/source/optimizer/onnx/../TransformGroupConvolution.cpp` (both guards) — correct.
- `tools/converter/source/optimizer/onnxextra/OnnxConvolutionMerge.cpp` — the locals are pre-initialized (`group = 1`, `dilation_h/w = 1`, `stride_h/w = 1` at lines 54-58), so the guard only rejects genuinely malformed attributes. A `nullptr` op is already handled gracefully at `OnnxExtraManager.cpp:70` (`MNN_ERROR` + `return false`).
- `source/backend/cpu/compute/ConvolutionFloatFactory.cpp` — fine.

## Blocking issue 1: `ConvolutionTflite.cpp` logs but does not abort

```cpp
if (ci <= 0) { MNN_ERROR("Conv2D weight has invalid input channel:%d\n", ci); }
if (inputShape.size() == 4 && ci > 0 && inputShape[3] > ci) { group = inputShape[3] / ci; }
```

Conversion continues after the error and can emit an invalid MNN model (`inputCount=0`, empty weight). This file already has a convention for fatal conversion errors — see `ConvolutionTflite.cpp:186` and `:279`, which are handled in `liteConverter.cpp:384-387` (`if (op->type == MNN::OpType_MAX) { MNNNetT.reset(); return 0; }`). Suggest:

```cpp
if (ci <= 0) {
    MNN_ERROR("Conv2D weight has invalid input channel:%d\n", ci);
    dstOp->type = MNN::OpType_MAX;
    return;
}
```

## Blocking issue 2: the `ShapePool.cpp` guard rejects *valid* ONNX models

Unlike `Convolution2DCommon`, the `Pool` table in `schema/default/CaffeOp.fbs` has **no default** for `strideX`/`strideY` — the generated `PoolT()` initializes them to `0` (vs `Convolution2DCommonT()` which initializes `strideX(1)`). And `OnnxPooling.cpp:105-160` only assigns stride *inside* the `"strides"` attribute branch. But `strides` is an **optional** attribute in the ONNX spec (default 1 per spatial axis), so a perfectly valid model can reach shape inference with `strideX == 0`.

I built two `onnx.checker`-valid MaxPool models and converted/ran them with the in-tree tools (`build/MNNConvert`, `build/MNNDump2Json`, `build/MNNV2Basic.out`):

| model (both onnx.checker-valid) | convert | stride in .mnn | inference |
|---|---|---|---|
| explicit `strides=[1,1]` | Converted Success | `strideX=1, strideY=1` | flops 0.002357 M → 1x8x7x7 (correct) |
| **`strides` omitted** | Converted Success | **`strideX=0, strideY=0`** | flops 0.000526 M → **1x8x1x1 (silently wrong)** |

Note there is no crash on arm64 today: AArch64 `sdiv` returns 0 on divide-by-zero instead of trapping, so the bug currently manifests as a silently wrong output shape. (I'd expect SIGFPE on x86_64, but I did not verify that.)

So the underlying bug is real, but with this PR the same valid model would be **hard-rejected** at shape inference instead. The guard needs to be paired with proper defaulting:

- `OnnxPooling.cpp` — initialize `strideX = strideY = 1` before the attribute loop.
- `PoolTorch.cpp:32-40` — when `inputs.size() <= 2`, PyTorch's `max_pool2d` defaults stride to `kernel_size`, so use the kernel size there rather than leaving it at 0.

With those in place the `ShapePool.cpp` guard becomes a true safety net rather than a regression.

## Minor follow-up (out of scope, just flagging)

`Pool3D` has the same class of issue and isn't covered here: `OnnxPooling.cpp:18-56` can leave `pool->strides` empty, and `ShapePool3D.cpp:53` dereferences `(*layer->strides())[i]` with no null/size check. Happy to see it in a separate PR.

Also, regarding the 380 passing unit tests — they don't exercise this scenario, since no existing test model omits the pooling `strides` attribute. A converter test case for ONNX pooling default attributes would be a valuable addition.

Thanks again — once the two points above are addressed I think this is good to go.

JingliangGao:
Thanks for the thorough review, and for taking the time to test this empirically — the stride-omitted MaxPool case was definitely one I had missed. Both points are now addressed in b0e1438:

**Blocking issue 1 (`ConvolutionTflite.cpp`):** fixed as suggested — the `ci <= 0` branch now sets `dstOp->type = MNN::OpType_MAX` and returns, so the conversion aborts cleanly through the existing `liteConverter.cpp` path instead of continuing and emitting an invalid model.

**Blocking issue 2 (pool stride defaults):** agreed, a hard reject was the wrong call there. I kept the `ShapePool.cpp` guard as a safety net and added the proper defaulting:
- `OnnxPooling.cpp`: `strideX`/`strideY` are now initialized to 1 before the attribute loop, matching the ONNX spec default (1 per spatial axis).
- `PoolTorch.cpp`: when no stride input is provided, stride now defaults to the kernel size, matching PyTorch `max_pool2d`/`avg_pool2d` behavior.

Verified locally after rebuilding `MNNConvert` / `MNNDump2Json` / `MNNV2Basic.out`: both of my `onnx.checker`-valid MaxPool test models (explicit `strides=[1,1]` and `strides` omitted) now convert with `strideX=1, strideY=1` and run with 0.002357 M flops each — matching your "correct" reference. I also re-ran the original malformed models from #4797: all of them now fail with clear error messages instead of crashing, and the tflite case aborts cleanly with `Conv2D weight has invalid input channel:0`.

**Pool3D:** thanks for flagging it — I will open a separate PR covering the `OnnxPooling.cpp` / `ShapePool3D.cpp` gap, and add a converter test case for ONNX pooling default attributes there.

JingliangGao:
Hi @wangzhaode, just checking in — have you had a chance to take another look at the updates in b0e1438? Hope the changes address your two points, but if anything still looks off or isn't quite what you had in mind, just let me know and I'll be happy to adjust. Thanks again for your time!

JingliangGao:
No answer. Just close it.

wangzhaode:
Thanks for addressing both blocking points in b0e1438, and sorry for the delayed follow-up. The updated direction looks right. Could you please reopen this PR? We will continue the final review and validation from the latest commit.

JingliangGao:
> Thanks for addressing both blocking points in [b0e1438](https://github.com/alibaba/MNN/commit/b0e14384691539f5b11f74b1ce245074477b8692), and sorry for the delayed follow-up. The updated direction looks right. Could you please reopen this PR? We will continue the final review and validation from the latest commit.

@wangzhaode  Ok,  I have reopened this PR.  Please review this PR again when you have time.

wangzhaode (approved):

PR #4823, opened 2026-09-01
[Converter:Bugfix] Fix MNNConvert crash on malformed tflite models (#4796)

### LINKED ISSUE #4796: [bug] MNNConvert dereferences null builtin_options and empty shape vectors on a malformed tflite model
# MNNConvert dereferences null builtin_options and empty shape vectors on a malformed tflite model

[attachments_mnn_tflite_options.zip](https://github.com/user-attachments/files/31326599/attachments_mnn_tflite_options.zip)

Attach: 11_tflite_conv_weight_shape_rank.bin (2,476), 13_tflite_conv_options_null.bin (2,476), 16_tflite_conv_bias_copy_overflow.bin (2,476), 20_tflite_pooling_options_null.bin (798), 22_tflite_weight_reorder_write.bin (2,476), 29_tflite_op_creator_dispatch.bin (2,476), 41_shape_concat_axis.bin (2,476), 43_tflite_depthwise_weight_shape.bin (782)

----------------------------------------------------------------------

`MNNConvert -f TFLITE` crashes on eight small malformed models, because the tflite converter takes the operator options union, the tensor shapes and the operator input lists straight from the file and uses them without checking, and the `DCHECK` guards that look like they check them only print a line and fall through.

All eight are the same root cause — a value from the model used without a check that stops — and can be fixed together. The reproducers and the full sanitizer output for each are attached.

| crash site | function | what is not checked | sanitizer | reproducer |
|---|---|---|---|---|
| `tools/converter/source/tflite/ConvolutionTflite.cpp:64` | `Conv2DTflite::run` | `weightShape.size()`; the `DCHECK` at :63 does not stop | SEGV, READ at 0x0 | `11_tflite_conv_weight_shape_rank.bin` (2,476 B) |
| `tools/converter/source/tflite/ConvolutionTflite.cpp:272` | `Conv2DTflite::run` | `AsConv2DOptions()` at :37 returns null | SEGV, READ at 0xc | `13_tflite_conv_options_null.bin` (2,476 B) |
| `tools/converter/source/tflite/ConvolutionTflite.cpp:327` | `Conv2DTflite::run` | length of the bias buffer read by `memcpy` | SEGV, READ at 0x4 | `16_tflite_conv_bias_copy_overflow.bin` (2,476 B) |
| `tools/converter/source/tflite/ConvolutionTflite.cpp:451` | `FullConnectedTflite::run` | `AsFullyConnectedOptions()` at :447 returns null | SEGV, READ at 0x8 | `29_tflite_op_creator_dispatch.bin` (2,476 B) |
| `tools/converter/source/tflite/PoolingTflite.cpp:74` | `PoolingTflite::run` | `AsPool2DOptions()` at :30 returns null | SEGV, READ at 0x14 | `20_tflite_pooling_options_null.bin` (798 B) |
| `tools/converter/source/tflite/DepthwiseConv2DTflite.cpp:83` | `DepthwiseConv2DTflite::run` | `weightShape.size()`; the `DCHECK` at :81 does not stop | SEGV, READ at 0x4 | `43_tflite_depthwise_weight_shape.bin` (782 B) |
| `tools/converter/source/tflite/TfliteUtils.cpp:107` | `convertDataFormatTflite` | `weightSize` (ConvolutionTflite.cpp:68) against the four loop bounds | SEGV, WRITE at 0x0 | `22_tflite_weight_reorder_write.bin` (2,476 B) |
| `source/shape/ShapeConcat.cpp:62` | `ConcatSizeComputer::onComputeSize` | `inputs` may be empty; the assert at :17 is commented out | SEGV, READ at 0x0 | `41_shape_concat_axis.bin` (2,476 B) |

## error log

`13_tflite_conv_options_null.bin`, a CONV_2D operator whose `builtin_options` union holds something other than `Conv2DOptions` (argument lists elided, full output attached):

```
==221==ERROR: AddressSanitizer: SEGV on unknown address 0x00000000000c (pc 0x7f465e129dc7 bp 0x7ffe287a54e0 sp 0x7ffe287a5220 T0)
==221==The signal is caused by a READ memory access.
==221==Hint: address points to the zero page.
    #0 0x7f465e129dc7 in Conv2DTflite::run(...) /w/src/tools/converter/source/tflite/ConvolutionTflite.cpp:272:54
    #1 0x7f465e173a5d in tflite2MNNNet(...) /w/src/tools/converter/source/tflite/liteConverter.cpp:383:22
    #2 0x7f465ce8c416 in MNN::Cli::convertModel(modelConfig&) /w/src/tools/converter/source/common/cli.cpp:644:24
    #3 0x55ec3562ab71 in main /w/src/tools/converter/source/MNNConverter.cpp:20:12

SUMMARY: AddressSanitizer: SEGV /w/src/tools/converter/source/tflite/ConvolutionTflite.cpp:272:54 in Conv2DTflite::run(...)
```

`22_tflite_weight_reorder_write.bin` is the one write of the eight:

```
==976==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x7f17951977e1 bp 0x7ffeea9acb60 sp 0x7ffeea9aca40 T0)
==976==The signal is caused by a WRITE memory access.
==976==Hint: address points to the zero page.
    #0 0x7f17951977e1 in convertDataFormatTflite(float const*, float*, int, int, int, int, bool) /w/src/tools/converter/source/tflite/TfliteUtils.cpp:107:64
    #1 0x7f179515c3a1 in Conv2DTflite::run(...) /w/src/tools/converter/source/tflite/ConvolutionTflite.cpp:309:17
    #2 0x7f17951a5a5d in tflite2MNNNet(...) /w/src/tools/converter/source/tflite/liteConverter.cpp:383:22
    #3 0x7f1793ebe416 in MNN::Cli::convertModel(modelConfig&) /w/src/tools/converter/source/common/cli.cpp:644:24

SUMMARY: AddressSanitizer: SEGV /w/src/tools/converter/source/tflite/TfliteUtils.cpp:107:64 in convertDataFormatTflite(float const*, float*, int, int, int, int, bool)
```

## context

- MNN at commit `19af6dab54a8a238c1e684bf5a9584f1f283f5cc`, from `https://github.com/alibaba/MNN/archive/19af6da.tar.gz`, sha256 `49c4a56f22abb749f99bc2f77ab8e9a1d047258746ee725ad74771ce05a682d5`. cmake reports MNN version 3.6.1. Every file quoted below is byte-identical to the one in that archive.
- Build: `cmake -DCMAKE_BUILD_TYPE=Release -DMNN_BUILD_CONVERTER=ON`, everything else left at its default, `CMAKE_CXX_FLAGS` empty. Two builds were made from the same options: one with clang 16.0.6 through AFL++'s `afl-clang-fast++` with `AFL_USE_ASAN=1`, which produced the output above, and one with `/usr/bin/clang++-16` and no sanitizer and no coverage.
- Platform: x86_64 Linux.

## how to reproduce

1. Unpack the archive above and build the converter: `cmake -G Ninja -DCMAKE_BUILD_TYPE=Release -DMNN_BUILD_CONVERTER=ON <source dir> && ninja MNNConvert`.
2. Run the converter on one of the attached files, for example:

   ```
   ./MNNConvert -f TFLITE --modelFile 13_tflite_conv_options_null.bin --MNNModel /dev/null --bizCode check
   ```
3. Repeat for the other seven files in the table. Each of the eight crashes on five runs out of five on the build with no sanitizer and no coverage.

## more

**1. `AsXxxOptions()` returns null when the union tag does not match, and the result is used without a test.** The accessor in the checked-in generated header, `tools/converter/source/tflite/schema/schema_generated.h:2764`:

```cpp
  Conv2DOptionsT *AsConv2DOptions() {
    return type == BuiltinOptions_Conv2DOptions ?
      reinterpret_cast<Conv2DOptionsT *>(value) : nullptr;
  }
```

`ConvolutionTflite.cpp:37` stores that result and `:272` dereferences it:

```cpp
    const auto& tfliteConvOption = tfliteOp->builtin_options.AsConv2DOptions();
...
        const auto acticationFun = tfliteConvOption->fused_activation_function;
```

The same shape is at `PoolingTflite.cpp:30` used at `:74`, and `ConvolutionTflite.cpp:447` used at `:451`. A CONV_2D operator carrying, say, `Pool2DOptions` is enough. The fault addresses are the member offsets read through the null pointer: 0xc is `fused_activation_function` in `Conv2DOptionsT`, 0x14 is the same field in `Pool2DOptionsT`, and 0x8 is `keep_num_dims` in `FullyConnectedOptionsT`.

**2. `DCHECK` prints and continues.** `tools/converter/include/logkit.h:69` and `:81`:

```cpp
#define CHECK(x) \
    if (!(x))    \
LogMessage(__FILE__, __LINE__).stream() << "Check failed: " #x << " ==> "
...
#define DCHECK(x) CHECK(x)
```

There is no abort in `LogMessage` (`logkit.h:139`); its destructor writes a newline. So `ConvolutionTflite.cpp:63-67` runs on to the subscripts whatever the check said:

```cpp
    DCHECK(weightShape.size() == 4) << "Conv2D weight ERROR!";
    const int co         = weightShape[0];
    const int kh         = weightShape[1];
    const int kw         = weightShape[2];
    const int ci         = weightShape[3];
```

With an empty `shape` vector the element load faults at address 0x0 here, and at 0x4 at `DepthwiseConv2DTflite.cpp:83`, which reads `weightShape[1]` after the identical `DCHECK` at `:81`.

**3. Lengths taken from the model are never compared with the buffers they index.** `ConvolutionTflite.cpp:298` does check the weight buffer before use, but the bias path two dozen lines later does not, `:323-327`:

```cpp
        std::vector<float> biasData(co, 0.0f);
        if (inputSize == 3) {
            const auto& biasTensor = tfliteTensors[tfliteOp->inputs[2]];
            auto biasDataPtr       = reinterpret_cast<const float*>(tfliteModelBuffer[biasTensor->buffer]->data.data());
            ::memcpy(biasData.data(), biasDataPtr, sizeof(float) * co);
```

The destination is sized `co`, so it is the source that is unchecked: an empty bias buffer gives `data()` == null and the copy reads from the zero page, which is the SEGV at 0x4.

The write at `TfliteUtils.cpp:107` is the same mismatch one step further on. `ConvolutionTflite.cpp:68` computes `const int weightSize = co * kh * kw * ci;`, `:303-304` size the destination from that single `int`, and the loops in `convertDataFormatTflite` then walk the four dimensions independently:

```cpp
bool convertDataFormatTflite(const float* src, float* dst, int KH, int KW, int CI, int CO, bool deconv) {
    DCHECK(KH > 0);
...
    for (int oc = 0; oc < CO; ++oc) {
        for (int ic = 0; ic < CI; ++ic) {
            for (int h = 0; h < KH; ++h) {
                for (int w = 0; w < KW; ++w) {
                    dst[(oc * CI + ic) * KH * KW + h * KW + w] = deconv ? ...
```

When the product is zero or wraps to zero in `int` while the four bounds are not, the destination vector is empty, `dst` is null, and the first store writes to address 0x0 — the WRITE in the second log. The `DCHECK`s at the top of the function are the same no-ops as above.

**4. `ShapeConcat.cpp` reads `inputs[0]` from a vector that may be empty.** The guard on line 17 is commented out and line 62 indexes anyway:

```cpp
        MNN_ASSERT(1 == outputs.size());
        // MNN_ASSERT(inputs.size() >= 2);
...
        TensorUtils::getDescribe(outputs[0])->dimensionFormat = TensorUtils::getDescribe(inputs[0])->dimensionFormat;
```

The sanitizer column, 62:90, is the `inputs[0]` on the right-hand side. This one is reached after conversion, from the converter's own optimizer pass (`optimizer/merge/TensorConverterMerge.cpp:70` → `Expr::requireInfo` → `Executor::computeInfo`), so a Concat op with no inputs survives conversion and faults in shape computation.

### suggested fix

The file already has an idiomatic way to reject an operator: set `dstOp->type = MNN::OpType_MAX` and return, which `liteConverter.cpp:384` turns into a clean conversion failure. Applying that at the three sites in `ConvolutionTflite.cpp`:

```diff
--- a/tools/converter/source/tflite/ConvolutionTflite.cpp
+++ b/tools/converter/source/tflite/ConvolutionTflite.cpp
@@
     const auto& tfliteConvOption = tfliteOp->builtin_options.AsConv2DOptions();
+    if (nullptr == tfliteConvOption) {
+        DLOG(ERROR) << "CONV_2D operator carries no Conv2DOptions";
+        dstOp->type = MNN::OpType_MAX;
+        return;
+    }
     const int inputIndex     = tfliteOp->inputs[0];
@@
     const auto& weightShape = weightTensor->shape;
-    DCHECK(weightShape.size() == 4) << "Conv2D weight ERROR!";
+    if (4 != weightShape.size()) {
+        DLOG(ERROR) << "Conv2D weight shape is not 4-D";
+        dstOp->type = MNN::OpType_MAX;
+        return;
+    }
     const int co         = weightShape[0];
@@
         std::vector<float> biasData(co, 0.0f);
         if (inputSize == 3) {
             const auto& biasTensor = tfliteTensors[tfliteOp->inputs[2]];
-            auto biasDataPtr       = reinterpret_cast<const float*>(tfliteModelBuffer[biasTensor->buffer]->data.data());
-            ::memcpy(biasData.data(), biasDataPtr, sizeof(float) * co);
+            const auto& biasBuffer = tfliteModelBuffer[biasTensor->buffer]->data;
+            if (biasBuffer.size() < sizeof(float) * (size_t)co) {
+                DLOG(ERROR) << "Conv2D bias buffer is shorter than the output count";
+                dstOp->type = MNN::OpType_MAX;
+                return;
+            }
+            ::memcpy(biasData.data(), biasBuffer.data(), sizeof(float) * co);
         }
```

The same three lines cover `PoolingTflite.cpp:30`, `ConvolutionTflite.cpp:447` and `DepthwiseConv2DTflite.cpp:87` for the null options, and `DepthwiseConv2DTflite.cpp:81` for the rank. The four dimensions want a positive-value check before line 68 multiplies them, so that `weightSize` cannot be zero or wrap while the loop bounds are not — that closes the write as well as the read. For `ShapeConcat.cpp`, restoring line 17 as a real early `return false` when `inputs` is empty is enough.

Turning `DCHECK` into something that stops would fix a much wider class of these at once, but it changes behaviour across the whole converter, so the per-site checks above are the smaller change.

fshuang8299:
### Summary
`MNNConvert -f TFLITE` segfaults on eight malformed models reported in #4796 (all reproducers + ASAN traces attached there). Every crash is a value taken straight from the model file and used without a check that stops execution.

### Root causes
1. **Null `builtin_options` union**: `AsConv2DOptions()` / `AsPool2DOptions()` / `AsDepthwiseConv2DOptions()` / `AsTransposeConvOptions()` / `AsFullyConnectedOptions()` return `nullptr` when the union tag does not match, and the result was dereferenced unconditionally (e.g. `fused_activation_function` read through null → SEGV at offset 0xc/0x8/0x14).
2. **`DCHECK` is log-and-continue**: `logkit.h` defines `DCHECK(x)` as `CHECK(x)` whose body only prints (no abort), so `DCHECK(weightShape.size() == 4)` fell through into `weightShape[0..3]` element loads on empty shapes.
3. **Unchecked arithmetic**: `co * kh * kw * ci` in `int` wraps for adversarial dims (product of large-but-positive dims wraps to a small positive value), producing an undersized destination followed by an out-of-bounds write inside `convertDataFormatTflite` (the WRITE SEGV at TfliteUtils.cpp:107).
4. **Unchecked bias buffer**: `::memcpy(biasData.data(), biasDataPtr, sizeof(float) * co)` reads from the zero page / OOB when the bias buffer is empty or undersized.
5. **`ConcatSizeComputer`**: dereferenced `inputs[0]` on an empty input list (assert commented out) and trusted a rank that can overflow the 9-slot dim array.

### Fix
Reject the malformed operator using the converter's idiomatic `dstOp->type = MNN::OpType_MAX` pattern (`liteConverter.cpp:384` turns it into a clean conversion failure with exit code 1), and fail closed everywhere:
- Null-check every `builtin_options` accessor (Conv2D, DepthwiseConv2D, TransposeConv, Pool2D, FullyConnected).
- Validate weight shape is 4-D, all dims positive, and `weightSize` computed in `int64` does not overflow `int` — before any allocation or division (also protects the `inputShape[3] / ci` group calc).
- Bound-check every bias buffer with `size() >= sizeof(T) * count` before `memcpy` / vector range-construction (float, INT8 and UINT8 paths).
- `convertDataFormatTflite` now validates dims and src/dst and returns `false` instead of looping through null pointers (defense in depth for all 4 call sites).
- `ConcatSizeComputer` rejects empty inputs, invalid rank (> `MNN_MAX_TENSOR_DIM`), and out-of-range axis (per input).

### Verification
- **All 8 reproducers from #4796**: 8/8 SEGV on master → 8/8 clean exit after fix (exit 1 for rejected models).
- **AddressSanitizer** build: all 8 reproducers + the legal model run with zero ASAN reports.
- **No regression**: a generated legal tflite model (conv + depthwise + pool + fc + concat, FLOAT32) converts identically to master — op list byte-for-byte identical, only the random model UUID differs.

Fixes #4796


fshuang8299:
## Verification Report

Complete verification of the fix against the 8 reproducers from #4796, plus regression checks. All tests run locally on x86_64 Linux.

### 1. Crash reproducers (functional)

All 8 files attached to #4796, run with `MNNConvert -f TFLITE --modelFile <file> --MNNModel /dev/null --bizCode check`:

| Master (baseline) | This PR |
|---|---|
| 8/8 SIGSEGV (exit 139) | 8/8 clean exit (exit 1 for rejected models) |

### 2. No-regression: legal model conversion

Generated a legal FLOAT32 tflite model (CONV_2D → DEPTHWISE_CONV_2D → AVERAGE_POOL_2D → CONCATENATION, 8 tensors). Converted with master baseline and with this PR:

- Conversion succeeds on both
- Output op list identical (same op types / names / input-output indexes)
- **Numeric inference comparison**: both converted models run on CPU with the same fixed input → **128 output elements bit-identical (max abs diff = 0)**

### 3. Quantized path (UINT8)

Generated a UINT8-quantized CONV_2D model (with INT32 bias + quantization params). Conversion succeeds identically on baseline and this PR.

### 4. Memory safety (AddressSanitizer)

ASAN build (`-fsanitize=address`): all 8 reproducers + the legal model run with **zero ASAN reports**. The only LeakSanitizer output is a pre-existing 8-byte static registration object (`onnxOpConverterRegister<UnaryOnnx>`), present on every MNNConvert invocation regardless of this change.

### 5. MNN test suite

`run_test.out` (CPU backend, Precision_High): **380/380 passed, 0 failed** — covers the touched `source/shape/ShapeConcat.cpp` (runtime shape computation) and general core regressions.

### 6. Style checks (matching repo CI)

- Commit message: `[Converter:Bugfix] Fix MNNConvert crash on malformed tflite models (#4796)` — matches the CI `Check Commit Message Format` rule
- Code format: checked with `git clang-format --diff origin/master` using clang-format **17.0.6** (the version the CI `Check Changed Lines Format` job installs) — **0 issues**

### Notes on policy choices

- **Depthwise INT8 empty/undersized weight → reject the op** (`OpType_MAX`), unlike Conv2D INT8 which skips the weight. The depthwise path assumes a constant weight buffer; without it the op cannot be represented. Rejecting is fail-closed and avoids silently emitting a zero-weight model.
- **ConcatSizeComputer rejects rank-0 inputs and out-of-range axis** instead of indexing `dim[axis]` blindly (the old path was itself out of bounds for scalar inputs / bad axis).
- Rejected ops use the converter's existing idiom (`dstOp->type = MNN::OpType_MAX`), which `liteConverter.cpp:384` turns into a clean conversion failure (exit code 1) rather than a crash.

wangzhaode:
Thanks for fixing these TFLite converter crashes and for sharing the validation results! The direction looks good. Could you please make three small updates before we merge?

1. When a supplied bias buffer is too small, please reject the operator with OpType_MAX and return, rather than logging "ignore bias" and continuing. Continuing with zero or missing bias can produce a successfully converted model with incorrect results. Please apply this consistently to the Conv2D, DepthwiseConv2D and TransposeConv paths touched here.
2. In ShapeConcat.cpp, the first loop checks the rank and then immediately breaks, so only the first input gets the full rank validation. Please validate every input's rank before accessing its dimensions, including the upper bound and consistency with the output rank.
3. The int64_t weight-size multiplication can itself overflow for malformed dimensions before the subsequent check runs. Since the supported result must fit in int, please check the running product against INT_MAX / dim before each multiplication, then multiply. Please use this for both convolution and depthwise weight-size checks.

A few focused cases for truncated bias, an invalid later Concat input, and overflowing dimensions are sufficient; there is no need for a large new test suite. Thanks!

fshuang8299:
Thanks for the review. All three points are addressed, on top of current master (rebased onto a03b005).

**1. Bias buffer too small -> reject instead of "ignore bias"**

All seven sites in the paths you mentioned now fail the operator with `OpType_MAX` and return, instead of logging and continuing:

| path | location |
|---|---|
| CONV_2D UINT8 | `ConvolutionTflite.cpp:205` |
| CONV_2D INT8 | `ConvolutionTflite.cpp:307` |
| CONV_2D FLOAT | `ConvolutionTflite.cpp:378` |
| TRANSPOSE_CONV | `ConvolutionTflite.cpp:474` |
| DEPTHWISE_CONV_2D INT8 | `DepthwiseConv2DTflite.cpp:161` |
| DEPTHWISE_CONV_2D UINT8 | `DepthwiseConv2DTflite.cpp:237` |
| DEPTHWISE_CONV_2D FLOAT | `DepthwiseConv2DTflite.cpp:266` |

A null / empty bias buffer is rejected as well, not only a short one. No `ignore bias` occurrence is left in `tools/`.

**2. ShapeConcat: validate every input's rank**

The loop now validates each input's rank - including the `MNN_MAX_TENSOR_DIM` upper bound - before touching that input's dimensions, and requires it to match the output rank (`ShapeConcat.cpp:37-61`). The previous loop checked the rank and then `break`-ed, so only the first input was validated; a later input with a larger rank could index past the copied dims, and the comparison read `ob.dim[i]` that had never been initialised.

**3. Check the running product against INT_MAX / dim before multiplying**

`computeTfliteWeightSize()` (`TfliteUtils.hpp:63`, `TfliteUtils.cpp:118`) compares the running product with `INT_MAX / dim` before each multiplication, with `dim <= 0` rejected first so the division can never be by zero. Both the convolution and the depthwise weight-size checks use it (`ConvolutionTflite.cpp:98`, `:447`; `DepthwiseConv2DTflite.cpp:98`).

One note on this point: #4830 has since added `validTfliteShape()` (`liteConverter.cpp:31-49`), which applies the same `INT32_MAX / d` guard globally to every tensor, so for the overflow case the per-op check is now redundant end-to-end. I kept it since you asked for it and it keeps the op-level code self-contained, but I'm happy to drop it if you would rather rely on the global pre-pass.

**Behaviour changes**

- A CONV_2D / DEPTHWISE_CONV_2D / TRANSPOSE_CONV whose bias input is present but holds fewer than `sizeof(T) * count` bytes now fails conversion instead of converting with a silently zeroed bias. Ops without a bias input are unaffected (the bias stays optional).
- Concat with inputs of unequal rank now fails instead of reading uninitialised output dims. Concat with equal ranks is unchanged.
- Nothing else: 267 valid tflite models convert byte-identically to master (model UUID normalised).

**Verified after the rebase** (clang-format 17.0.6 CI check reports no diff, Release and ASan builds clean)

- 8 malformed models from #4796: exit cleanly, no SIGSEGV, AddressSanitizer-clean apart from a pre-existing 8-byte ONNX op-registry static-init leak that is unrelated to tflite.
- 267 valid tflite models: byte-identical to master.
- Repo test suite `run_test.out`: 380 passed / 0 failed.

**Focused cases**

1. Truncated bias buffer -> `CONV_2D bias buffer needs 12 bytes, got 0` (previously: converted with a zeroed bias). Note that with current master this specific file is now rejected even earlier by the flatbuffer verifier added in #4830.
2. A later Concat input with a mismatched rank -> `Concat op input 1 rank 3 not match output rank 2` (previously: `Error for concat size of op [  ], the 2 input not match output`, which read never-initialised output dims).
3. Weight dimensions whose product overflows -> `CONV_2D weight size overflow: 262156x524296x524288x655360` (previously the `int64_t` product wrapped, i.e. undefined behaviour; this file is also caught earlier now by `validTfliteShape`).

One more thing that came out of this: the new early returns made an existing `return 0` path in `liteConverter.cpp` leak the operator it had already built, and the pre-existing unsupported-op returns had the same problem. That is fixed minimally there by holding the `MNN::OpT` in a `unique_ptr` until it is appended to `oplists`.

wangzhaode:
Thanks for addressing the three review points and for the detailed follow-up validation! Sorry we did not respond sooner.

The branch currently has merge conflicts with master. Could you please rebase and resolve them, preserving the bias rejection, per-input Concat rank checks, and checked weight-size multiplication? We are not asking you to redo those changes or expand the scope.

Once the conflicts are resolved, we can do the final review of the rebased version. A focused build/smoke check around the conflict resolutions is enough; no larger test suite is needed. Thanks again!

fshuang8299:
@wangzhaode Done — rebased onto current master (451006a); one conflict in `ConvolutionTflite.cpp`.

All three items you asked to preserve are intact:

- **bias rejection** — unchanged
- **per-input Concat rank checks** — unchanged (`ShapeConcat.cpp` applied cleanly)
- **checked weight-size multiplication** — unchanged

The conflict was #4809's `ci <= 0` guard, added right after `weightSize = co * kh * kw * ci`. The version here already covers that case: it rejects on `co/kh/kw/ci <= 0`, and checks the running product against `INT_MAX / dim` before each multiply. That is a strict superset, so nothing from #4809 is dropped.

Smoke check around the resolution: the 8 reproducers from #4796 all exit cleanly, and `run_test.out` is 387/387.

Ready for the final review whenever you are.

wangzhaode:
Thanks for rebasing, preserving the earlier fixes, and running the reproducers and regression tests!

There is one remaining issue in `tools/converter/source/tflite/ConvolutionTflite.cpp`, inside `TransposeConvTflite::run()`: when `inputSize == 4`, the bias tensor is still read from `tfliteOp->inputs[2]`. TFLite TRANSPOSE_CONV uses inputs `[output_shape, weights, input, bias]`, so the bias should come from `inputs[3]`; `inputs[2]` is the activation tensor, as the input-index mapping later in this function also shows.

The wrong index predates this PR, but the new buffer-size rejection makes it reject a valid four-input model when the activation tensor has no constant buffer. Could you please change the bias lookup to `inputs[3]`, keeping the buffer validation on the actual bias tensor?

One small valid FP32 TRANSPOSE_CONV case with a runtime input and a nonzero constant bias is enough to check successful conversion and that the bias is preserved. There is no need to expand the scope or add a large test suite. Thanks again!

fshuang8299:
@wangzhaode Fixed — the bias now comes from `inputs[3]`.

`TransposeConvTflite::run()` was reading it from operand 2. As you pointed out, operand 2 is the activation tensor — the very operand this function assigns to `dstOp->inputIndexes[0]` a few lines below — so when that input is a runtime input it has no constant buffer at all, and the buffer-size check added in this PR turned that into a rejection of a valid model:

```
TRANSPOSE_CONV bias buffer needs 4 bytes, got 0
```

The lookup is now `tfliteOp->inputs[3]`, and the buffer validation is unchanged — it simply runs against the tensor that is actually copied into `biasData`. I also spelled the operand order out in the comment above `inputSize`, so the payload/consumer split does not have to be re-derived by hand next time.

Checked with a legal FP32 TRANSPOSE_CONV model (runtime activation input, nonzero constant bias, 1x1 kernel, stride 1):

* conversion succeeds (exit 0) and the bias reaches the `.mnn` — weights `[1,1,1,1]`, bias `[0.5, 1.5, 2.5]`;
* running the converted model on the CPU backend reproduces `out = sum(in) + bias` exactly: input `{1,1,2,2,3,3,4,4}` → `2.5 3.5 4.5 | 4.5 5.5 6.5 | 6.5 7.5 8.5 | 8.5 9.5 10.5`;
* for comparison, the same model was rejected before this change, and on master the same bias was silently dropped to `0`.

Nothing else was touched — one file, and the only functional line is the operand index. Regression re-run afterwards: 8/8 malformed reproducers still exit cleanly, `run_test.out` 387/387, and 1175 legal tflite models convert with an identical exit-code distribution.

wangzhaode (approved):

PR #4830, opened 2026-09-04
[Converter:Bugfix] Fail fast on invalid TFLite model and propagate optimizer failures

### LINKED ISSUE #4798: [bug] MNNConvert crashes in Program::create and the ONNX/TFLite extra-op transforms when conversion leaves null inputs
MNNConvert crashes in Program::create and the ONNX/TFLite extra-op transforms when conversion leaves null inputs

[attachments_mnn_expr_null.zip](https://github.com/user-attachments/files/31326613/attachments_mnn_expr_null.zip)

Attach: 03_tflite_program_create.bin (172 bytes)
        34_optimizer_subgraph_iterate.bin (640 bytes)
        35_expr_varp_copy_null.bin (544 bytes)
        39_tensor_describe_destroy.bin (768 bytes)
        40_core_alloc_align_size.bin (544 bytes)
        49_onnx_extra_type_null.bin (527 bytes)
        50_onnx_expr_create.bin (527 bytes)
        52_onnx_pool_kernel_list.bin (1,393 bytes)
        53_onnx_varp_copy.bin (527 bytes)
        54_onnx_variable_expr.bin (512 bytes)
        58_onnx_node_output_missing.bin (527 bytes)

When a front end cannot resolve an operator's inputs, or produces a net with no operators at all, it prints a message and keeps going, and the optimizer stage then uses the null or missing object it left behind without checking it — ten of the eleven attached malformed models crash `MNNConvert` that way, and the eleventh is the same habit in the allocator.

## error log

A TFLite model that converts to zero ops. `optimizeNetImpl` returns `nullptr` and the caller passes it straight to `Program::create`; `0x40` is `offsetof(NetT, oplists)`:

```
Start to Convert Other Model Format To MNN Model..., target version: 3.6
Start to Optimize the MNN Net...
AddressSanitizer:DEADLYSIGNAL
==18==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000040 (pc 0x7f1678ac79f5 bp 0x7ffc2fa39c90 sp 0x7ffc2fa39ae0 T0)
==18==The signal is caused by a READ memory access.
==18==Hint: address points to the zero page.
    #0 std::vector<std::unique_ptr<MNN::OpT>>::size() const  stl_vector.h:919:66
    #1 MNN::Express::Program::create(...)  tools/converter/source/optimizer/Program.cpp:122:41
    #2 MNN::Express::Program::create(MNN::NetT const*, bool, bool)  tools/converter/source/optimizer/Program.cpp:101:12
    #3 optimizeNet(...)  tools/converter/source/optimizer/PostConverter.cpp:675:20
    #4 MNN::Cli::convertModel(modelConfig&)  tools/converter/source/common/cli.cpp:702:20
    #5 main  tools/converter/source/MNNConverter.cpp:20:12
```

An ONNX `Conv` whose three input names do not resolve. The converter logs all three and builds the op anyway; `OnnxConvolutionTransform::onExecute` then indexes the empty input vector:

```
Start to Convert Other Model Format To MNN Model..., target version: 3.6
[20:24:08] :46: ONNX Model ir version: 48
[20:24:08] :47: ONNX Model opset version: 13
[20:24:08] :149: Check it out ==> 000Conv_o00p00_0 has empty input, the index is 0
[20:24:08] :149: Check it out ==> 000Conv_o00p00_0 has empty input, the index is 1
[20:24:08] :149: Check it out ==> 000Conv_o00p00_0 has empty input, the index is 2
Start to Optimize the MNN Net...
AddressSanitizer:DEADLYSIGNAL
==187120==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x7f344e345ce0 bp 0x7fff866cd3d0 sp 0x7fff866cd0a0 T0)
==187120==The signal is caused by a READ memory access.
==187120==Hint: address points to the zero page.
    #0 std::__shared_ptr<MNN::Express::Variable, ...>::operator=(...)  shared_ptr_base.h:1153:69
    #1 std::shared_ptr<MNN::Express::Variable>::operator=(...)  shared_ptr.h:359:65
    #2 MNN::Express::VARP::VARP(MNN::Express::VARP const&)  include/MNN/expr/Expr.hpp:52:18
    #3 MNN::Express::OnnxConvolutionTransform::onExecute(EXPRP) const  tools/converter/source/optimizer/onnxextra/OnnxConvolutionMerge.cpp:143:29
    #4 MNN::Express::$_0::operator()() const::'lambda0'(EXPRP)::operator()(EXPRP) const  tools/converter/source/optimizer/onnxextra/OnnxExtraManager.cpp:69:37
    #9 MNN::Express::TemplateMerge::insertTemplate(...)::$_0::operator()(EXPRP) const  tools/converter/source/optimizer/TemplateMerge.cpp:197:16
    #14 MNN::Express::TemplateMerge::onExecute(...)  tools/converter/source/optimizer/TemplateMerge.cpp:149:21
    #16 MNN::Express::RunExtraPass(...)  tools/converter/source/optimizer/PostConverter.cpp:205:11
    #17 MNN::Express::optimizeNetImpl(...)  tools/converter/source/optimizer/PostConverter.cpp:330:24
    #22 optimizeNet(...)  tools/converter/source/optimizer/PostConverter.cpp:674:38
    #23 MNN::Cli::convertModel(modelConfig&)  tools/converter/source/common/cli.cpp:702:20
    #24 main  tools/converter/source/MNNConverter.cpp:20:12
```

Both logs are edited only by removing the frame addresses and the template parameter lists; the gaps in the second numbering are `std::function` adapter frames, and the trailing libc frames and `SUMMARY` line are cut. The unedited output goes with the reproducers on request.

## context

* MNN at commit `19af6da`, `tools/converter` (`MNNConvert`).
* Built with CMake, `-DCMAKE_BUILD_TYPE=Release -DMNN_BUILD_CONVERTER=ON`, clang 16, nothing else changed.
* x86_64 Linux, libstdc++ 11.
* `MNN_ASSERT` expands to nothing in this configuration: it is guarded by `DEBUG` (`include/MNN/MNNDefine.h:42-53`), and `CMakeLists.txt:408-409` defines `DEBUG` only when `CMAKE_BUILD_TYPE` is `Debug`. The `MNN_ASSERT` guards on these paths are not present in a Release build.
* The stacks and addresses above come from a second build of the same commit with AddressSanitizer enabled. The crashes themselves reproduce on the plain Release build, with no sanitizer and no coverage instrumentation.

## how to reproduce

1. Build `MNNConvert` from commit `19af6da` in the configuration above.
2. Run it on an attached reproducer, choosing the framework from the file name — `-f TFLITE` for the five `tflite`/optimizer/core files, `-f ONNX` for the six `onnx` files:

   ```
   MNNConvert -f TFLITE --modelFile 03_tflite_program_create.bin --MNNModel /dev/null --bizCode check
   MNNConvert -f ONNX   --modelFile 53_onnx_varp_copy.bin        --MNNModel /dev/null --bizCode check
   ```
3. The process crashes. Under AddressSanitizer ten of the eleven are reported as SEGV; `40_core_alloc_align_size.bin` is reported as an over-size allocation request instead.

Each of the eleven reproducers crashes on every run — 5 runs out of 5 on the plain Release build. The file:line below are the ones the AddressSanitizer build reports.

## more

The eleven crash sites, all downstream of the same unchecked state:

| reproducer | `-f` | crash site | fault |
|---|---|---|---|
| `03_tflite_program_create.bin` | TFLITE | `optimizer/Program.cpp:122` — `oplists.size()` on a null `NetT` | SEGV read @ `0x40` |
| `34_optimizer_subgraph_iterate.bin` | TFLITE | `optimizer/PostConverter.cpp:713` — `net->subgraphs` on a null `NetT` | SEGV read @ `0x98` |
| `35_expr_varp_copy_null.bin` | TFLITE | `optimizer/tflitextra/BinaryActivation.cpp:28` → `include/MNN/expr/Expr.hpp:52` | SEGV read @ `0x0` |
| `39_tensor_describe_destroy.bin` | TFLITE | `source/core/TensorUtils.hpp:86` — implicit `~NativeInsideDescribe` during `Program` teardown | SEGV read @ `0x9` |
| `40_core_alloc_align_size.bin` | TFLITE | `source/core/MNNMemoryUtils.cpp:24` — `malloc(size + …)` | request of `0x220000000048` bytes (34 TiB) |
| `49_onnx_extra_type_null.bin` | ONNX | `optimizer/onnxextra/OnnxExtraManager.cpp:57` — `main_as_Extra()->type()->str()` | SEGV read @ `0x619100009014` |
| `50_onnx_expr_create.bin` | ONNX | `express/Expr.cpp:240` — `op->main.AsBlob()->external` | SEGV read @ `0xb0` |
| `52_onnx_pool_kernel_list.bin` | ONNX | `optimizer/onnxextra/OnnxPooling.cpp:76` — `attr->list()->i()->size()` | SEGV read @ `0x6191000099dc` |
| `53_onnx_varp_copy.bin` | ONNX | `optimizer/onnxextra/OnnxConvolutionMerge.cpp:143` → `include/MNN/expr/Expr.hpp:52` | SEGV read @ `0x0` |
| `54_onnx_variable_expr.bin` | ONNX | `optimizer/merge/ConstantFolding.cpp:33` → `include/MNN/expr/Expr.hpp:123` | SEGV read @ `0x0` |
| `58_onnx_node_output_missing.bin` | ONNX | `onnx/onnxConverter.cpp:131` — `onnxNode.output(0)` on a node with no outputs | SEGV read @ `0x8` |

### the failure is not reported, so the next stage runs on nothing

There are three places where a failed conversion produces a value that means "nothing", and each is consumed unconditionally.

**1. A net with no ops becomes a null `NetT`.** `optimizer/PostConverter.cpp:278-280`, inside `optimizeNetImpl`:

```cpp
    if (originNet->oplists.size() <= 0) {
        return nullptr;
    }
```

Both callers use the result immediately. `optimizer/PostConverter.cpp:674-675`:

```cpp
    std::unique_ptr<MNN::NetT> net = ctx.RunOptimize(originNet, empty);
    auto program = Program::create(net.get(), true, true);
```

and `optimizer/PostConverter.cpp:705-713`:

```cpp
    net = ctx.RunOptimize(net, empty);
    ...
    for (auto& iter : net->subgraphs) {
```

`Program::create` takes `net->oplists` by reference and reads `.size()` at `optimizer/Program.cpp:122` — offset `0x40` in `NetT`, the address in the first log. The subgraph loop reads offset `0x98`, `NetT::subgraphs`. Reproducers 03 and 34. `common/cli.cpp:703` then dereferences the same return value again (`finalNet->extraTensorDescribe`) with no check.

**2. An unresolved ONNX input becomes index `-1`, and then a null `VARP` or a missing one.** `onnx/onnxConverter.cpp:146-155`:

```cpp
        for (int k = 0; k < onnxNode.input_size(); k++) {
            int inputIdx = scope->lookupTensor(onnxNode.input(k));
            if (inputIdx < 0) {
                LOG(INFO) << "Check it out ==> " << MNNOp->name << " has empty input, the index is " << k;
            }
            MNNOp->inputIndexes.push_back(inputIdx);
        }
        for (int k = onnxNode.input_size() - 1; k >= 0 && MNNOp->inputIndexes[k] < 0; --k) {
            MNNOp->inputIndexes.pop_back();
        }
```

The message is `LOG(INFO)` and conversion continues. A trailing `-1` is popped, so the op ends up with fewer inputs than its op type has; a `-1` in the middle survives and `optimizer/Program.cpp:41-43` turns it into a null `VARP`:

```cpp
        if (input < 0) { // optional input
            inputVars.emplace_back(nullptr);
            continue;
        }
```

The TFLite front end does the same by dropping the index outright, `tflite/liteConverter.cpp:371-375`:

```cpp
            for (int i = 0; i < ops[j]->inputs.size(); i++) {
                if (ops[j]->inputs[i] >= 0) {
                    op->inputIndexes.emplace_back(ops[j]->inputs[i]);
                }
            }
```

Every transform then indexes that vector without checking it first. `optimizer/onnxextra/OnnxConvolutionMerge.cpp:141-148` even does the indexing one line *above* its own size check:

```cpp
        auto inputs         = expr->inputs();
        const int inputSize = static_cast<int32_t>(inputs.size());
        auto x            = inputs[0];
        if (inputSize != 3 && inputSize != 2) {
            MNN_ERROR("Convolution Input ERROR!\n");
            return nullptr;
        }
        auto weight = inputs[1];
```

With `inputs` empty, `inputs[0]` reads through the vector's null data pointer, and the `VARP` copy constructor at `include/MNN/expr/Expr.hpp:52` faults at address 0. `optimizer/tflitextra/BinaryActivation.cpp:26-29` is the same shape behind an `MNN_ASSERT` that is compiled out:

```cpp
        auto inputs = expr->inputs();
        MNN_ASSERT(inputs.size() == 2);
        auto input0 = inputs[0];
        auto input1 = inputs[1];
```

Reproducers 35 and 53. Where the null `VARP` is present rather than missing, `optimizer/merge/ConstantFolding.cpp:32-33` calls a member function on it:

```cpp
        for (const VARP& input : expr->inputs()) {
            const Op* input_op = input->expr().first->get();
```

`Variable::expr()` reads `mFrom` through a null `this`, faulting at `include/MNN/expr/Expr.hpp:123`. Reproducer 54.

**3. An op whose parameter union was never filled.** `express/Expr.cpp:239-240`:

```cpp
    if (OpType_Const == op->type || OpType_TrainableParam == op->type) {
        if (!op->externalPath.empty() || (!op->main.AsBlob()->external.empty())) {
```

`AsBlob()` returns `nullptr` when `op->main.type != OpParameter_Blob`, which is what a half-built op has. `external` sits at offset `0xb0` in `BlobT`, the address in reproducer 50. `AsInput()` at line 231 and `AsBlob()` at line 251 are dereferenced the same way.

The remaining sites are unchecked flatbuffer accessors on the same half-built ops: `main_as_Extra()->type()` at `optimizer/onnxextra/OnnxExtraManager.cpp:57` (`engine()` on line 53 too), and `attr->list()->i()` at `optimizer/onnxextra/OnnxPooling.cpp:76` (`attr->key()` on line 74, `extraParam->type()` on line 84). `onnx/onnxConverter.cpp:131` reads `onnxNode.output(0)` before checking `output_size()`:

```cpp
        // name maybe null, use the first output name as node-name
        const auto& name = onnxNode.output(0);
```

Reproducer 39 is the same state reached on the teardown path: the `Program` built over a net that failed to convert is released at the end of `RunExtraPass`, `optimizer/PostConverter.cpp:219`, and the compiler-generated `~NativeInsideDescribe` (`source/core/TensorUtils.hpp:86`) faults releasing the `std::shared_ptr<TensorArrayAttr>` member declared at `source/core/TensorUtils.hpp:98`.

Reproducer 40 is separate but has the same shape: `source/core/MNNMemoryUtils.cpp:18-28` computes an allocation size straight from the model's shapes.

```cpp
extern "C" void *MNNMemoryAllocAlign(size_t size, size_t alignment) {
    MNN_ASSERT(size > 0);
...
    void **origin = (void **)malloc(size + sizeof(void *) + alignment);
```

The assert is compiled out, and the model asks for 34 TiB.

### suggested fix

Reject the model where the failure is first known, instead of letting it flow on. `onnx2MNNNet` already reports and returns `1` this way at `onnx/onnxConverter.cpp:37-38` and `52-53`:

```diff
--- a/tools/converter/source/onnx/onnxConverter.cpp
+++ b/tools/converter/source/onnx/onnxConverter.cpp
@@
+        if (onnxNode.output_size() <= 0) {
+            MNN_ERROR("[ERROR] Onnx node %d (%s) has no output\n", i, opType.c_str());
+            return 1;
+        }
         // name maybe null, use the first output name as node-name
         const auto& name = onnxNode.output(0);
@@
         for (int k = 0; k < onnxNode.input_size(); k++) {
             int inputIdx = scope->lookupTensor(onnxNode.input(k));
             if (inputIdx < 0) {
-                LOG(INFO) << "Check it out ==> " << MNNOp->name << " has empty input, the index is " << k;
+                MNN_ERROR("[ERROR] Onnx node %s input %d (%s) does not resolve\n",
+                          MNNOp->name.c_str(), k, onnxNode.input(k).c_str());
+                return 1;
             }
             MNNOp->inputIndexes.push_back(inputIdx);
         }
```

```diff
--- a/tools/converter/source/optimizer/PostConverter.cpp
+++ b/tools/converter/source/optimizer/PostConverter.cpp
@@
     std::unique_ptr<MNN::NetT> net = ctx.RunOptimize(originNet, empty);
+    if (nullptr == net) {
+        MNN_ERROR("Optimize failed: the model has no op\n");
+        return nullptr;
+    }
     auto program = Program::create(net.get(), true, true);
@@
     net = ctx.RunOptimize(net, empty);
+    if (nullptr == net) {
+        MNN_ERROR("Optimize failed: the model has no op\n");
+        return nullptr;
+    }
```

with the matching check on the return value in `common/cli.cpp:702`, which currently dereferences it on line 703.

For the transforms, one guard used by both extra-op managers is enough to cover the whole family, since `OnnxExtraManager.cpp:69` and `TFliteExtraManager.cpp:57` are the single dispatch points:

```cpp
// returns false if the op does not have `n` usable inputs
static bool hasInputs(EXPRP expr, int n) {
    const auto& inputs = expr->inputs();
    if ((int)inputs.size() < n) {
        return false;
    }
    for (int i = 0; i < n; ++i) {
        if (nullptr == inputs[i].get()) {
            return false;
        }
    }
    return true;
}
```

Independently of that, `OnnxConvolutionMerge.cpp:143` should simply move below its own size check:

```diff
         const int inputSize = static_cast<int32_t>(inputs.size());
-        auto x            = inputs[0];
         if (inputSize != 3 && inputSize != 2) {
             MNN_ERROR("Convolution Input ERROR!\n");
             return nullptr;
         }
+        auto x            = inputs[0];
         auto weight = inputs[1];
```

`Expr.cpp:240` needs the union tag checked before use:

```diff
     if (OpType_Const == op->type || OpType_TrainableParam == op->type) {
-        if (!op->externalPath.empty() || (!op->main.AsBlob()->external.empty())) {
+        auto blob = op->main.AsBlob();
+        if (nullptr == blob) {
+            MNN_ERROR("Const op %s has no blob parameter\n", op->name.c_str());
+            return nullptr;
+        }
+        if (!op->externalPath.empty() || (!blob->external.empty())) {
```

and the flatbuffer accessors at `OnnxExtraManager.cpp:53/57`, `OnnxPooling.cpp:74/76/84` need null checks on `engine()`, `type()`, `key()`, `list()` and `i()` before `->str()` / `->size()`.

Finally, `MNNMemoryAllocAlign` should refuse a size it cannot serve rather than relying on an assert that is not compiled in:

```diff
 extern "C" void *MNNMemoryAllocAlign(size_t size, size_t alignment) {
-    MNN_ASSERT(size > 0);
+    if (0 == size || size > SIZE_MAX - sizeof(void *) - alignment) {
+        return NULL;
+    }
```

though the real guard belongs earlier, where the tensor shape that produced 34 TiB is read from the model.

kabishou11:
## Summary

Narrowed per maintainer review on #4830 (rebased onto master after #4834).

**In scope (this PR):**
- `TfliteModel::readModel()` fail-fast: handle missing/empty files, reject unverified buffers without calling `UnPackModel` (root cause of #4798 SEGV when `LOG(FATAL)` does not abort).
- `tflite2MNNNet()` hard errors for null model, null subgraph/operator, invalid tensor shapes, empty `oplists` — reuses `tfliteAt<>()` from #4834 (no parallel index checks, no silent `continue`).
- `Program::createUnit` propagates `Expr::create` failure; `Program::create` returns `nullptr` so optimize/convert aborts cleanly instead of leaving holes in `varMap`.
- `optimizeNet` / `cli` null checks when program creation or optimization fails.

**Deferred to follow-up PRs (per review):**
- `express/Expr.cpp` null guards + `expr/create_guard` test
- `MNNMemoryUtils.cpp` overflow guard only (without `0 == size → NULL`)
- ONNX unresolved-input hard error (pending conversion regression)

## Test plan
- [ ] Malformed TFLite reproducers from #4798 fail with conversion errors (no SEGV)
- [ ] Converter build with `-DMNN_BUILD_CONVERTER=ON`

Fixes #4798 (converter path)

### DIFF
--- a/tools/converter/source/common/cli.cpp
+++ b/tools/converter/source/common/cli.cpp
@@ -700,6 +700,10 @@ bool Cli::convertModel(modelConfig& modelPath) {
     if (needOptimize) {
         std::cout << "Start to Optimize the MNN Net..." << std::endl;
         finalNet = optimizeNet(netT, modelPath.forTraining, modelPath, expectedPass);
+        if (nullptr == finalNet) {
+            MNN_ERROR("[ERROR] Optimize failed.\n");
+            return false;
+        }
         if (finalNet->extraTensorDescribe.size()>0 && expectedPass.empty()) {
             MNN_PRINT("MNN net has tensor quant info\n");
             computeUnaryBuffer(finalNet.get());
--- a/tools/converter/source/optimizer/PostConverter.cpp
+++ b/tools/converter/source/optimizer/PostConverter.cpp
@@ -599,6 +599,10 @@ bool fuseConstIntoSubgraph(MNN::NetT* net, const std::vector<MNN::SubGraphProtoT
         subnet->outputName = outputNames;
 
         std::unique_ptr<MNN::NetT> new_subnet = optimizeNetImpl(subnet, empty);
+        if (nullptr == new_subnet) {
+            MNN_ERROR("Optimize failed: subgraph %s has no op\n", mutable_subgraph->name.c_str());
+            return false;
+        }
         mutable_subgraph->nodes               = std::move(subnet->oplists);
 
         MNN::SubGraphProtoT* new_subgraph = mutable_subgraph;
@@ -672,7 +676,15 @@ std::unique_ptr<MNN::NetT> optimizeNet(std::unique_ptr<MNN::NetT>& originNet, bo
     // then subgraph depend on it may convert failed (nullptr) or wrong (error shape)
     // RunOptimize won't use subgraph, so we can do it before other subgraph optimize safely
     std::unique_ptr<MNN::NetT> net = ctx.RunOptimize(originNet, empty);
+    if (nullptr == net) {
+        MNN_ERROR("Optimize failed: the model has no op\n");
+        return nullptr;
+    }
     auto program = Program::create(net.get(), true, true);
+    if (nullptr == program) {
+        MNN_ERROR("Optimize failed: cannot create program\n");
+        return nullptr;
+    }
     auto addVars = [&](std::shared_ptr<Program> program, const std::vector<std::string>& tensorName) {
         for (const auto& iter : program->vars()) {
             if (iter.first < tensorName.size() && iter.first >= 0) {
@@ -691,6 +703,10 @@ std::unique_ptr<MNN::NetT> optimizeNet(std::unique_ptr<MNN::NetT>& originNet, bo
         CompleteSubGraph(inputs, ctx.subgraphs[idx]);
         auto new_graph = ctx.completed_subgraphs[idx];
         auto subProgram = Program::create(new_graph, true, true);
+        if (nullptr == subProgram) {
+            MNN_ERROR("Optimize failed: cannot create subgraph program\n");
+            return nullptr;
+        }
         subProgram->input(inputs, true);
         // add vars of subgraph, so inner subgraph can use them
         addVars(subProgram, new_graph->tensors);
@@ -703,6 +719,10 @@ std::unique_ptr<MNN::NetT> optimizeNet(std::unique_ptr<MNN::NetT>& originNet, bo
         CompleteSubGraph(inputs, subgraph);
     }
     net = ctx.RunOptimize(net, empty);
+    if (nullptr == net) {
+        MNN_ERROR("Optimize failed: the model has no op\n");
+        return nullptr;
+    }
 
     fuseConstIntoSubgraph(net.get(), ctx.completed_subgraphs);
     for (auto* subgraph : ctx.completed_subgraphs) {
--- a/tools/converter/source/optimizer/Program.cpp
+++ b/tools/converter/source/optimizer/Program.cpp
@@ -19,21 +19,21 @@ using namespace MNN;
 namespace MNN {
 namespace Express {
 
-void Program::createUnit(std::map<int, VARP>& varMap, std::vector<int>& inputIndexes, const std::vector<std::unique_ptr<OpT>>& oplists, MNN::OpT* op, const MNN::NetT* net, std::set<OpT*>& invalidSet, std::set<int>& extraInputIndexes) {
-    createUnit(varMap, inputIndexes, oplists, op, net->tensorName, invalidSet, extraInputIndexes, net);
+bool Program::createUnit(std::map<int, VARP>& varMap, std::vector<int>& inputIndexes, const std::vector<std::unique_ptr<OpT>>& oplists, MNN::OpT* op, const MNN::NetT* net, std::set<OpT*>& invalidSet, std::set<int>& extraInputIndexes) {
+    return createUnit(varMap, inputIndexes, oplists, op, net->tensorName, invalidSet, extraInputIndexes, net);
 }
 
-void Program::createUnit(std::map<int, VARP>& varMap, std::vector<int>& inputIndexes, const std::vector<std::unique_ptr<OpT>>& oplists,
+bool Program::createUnit(std::map<int, VARP>& varMap, std::vector<int>& inputIndexes, const std::vector<std::unique_ptr<OpT>>& oplists,
                     MNN::OpT* op, const std::vector<std::string>& tensorName, std::set<OpT*>& invalidSet, std::set<int>& extraInputIndexes, const MNN::NetT* net, std::map<std::string, int> TensorDescribeName) {
     if (invalidSet.find(op) != invalidSet.end()) {
-        return;
+        return true;
     }
     std::vector<VARP> inputVars;
     auto outputIndexes = op->outputIndexes;
     for (int j = 0; j < outputIndexes.size(); ++j) {
         if (varMap.find(outputIndexes[j]) != varMap.end()) {
             // Don't support multi op output to one index
-            return;
+            return false;
         }
     }
     invalidSet.insert(op);
@@ -46,7 +46,9 @@ void Program::createUnit(std::map<int, VARP>& varMap, std::vector<int>& inputInd
             for (int j = 0; j < oplists.size(); ++j) {
                 for (auto outputIndex : oplists[j]->outputIndexes) {
                     if (outputIndex == input) {
-                        createUnit(varMap, inputIndexes, oplists, oplists[j].get(), tensorName, invalidSet, extraInputIndexes, net, TensorDescribeName);
+                        if (!createUnit(varMap, inputIndexes, oplists, oplists[j].get(), tensorName, invalidSet, extraInputIndexes, net, TensorDescribeName)) {
+                            return false;
+                        }
                     }
                 }
             }
@@ -62,6 +64,10 @@ void Program::createUnit(std::map<int, VARP>& varMap, std::vector<int>& inputInd
         inputVars.emplace_back(varMap[input]);
     }
     auto expr = Expr::create(op, inputVars, outputIndexes.size());
+    if (nullptr == expr) {
+        MNN_ERROR("Create expr for op %s failed\n", op->name.c_str());
+        return false;
+    }
     expr->setName(op->name);
     for (int j = 0; j < outputIndexes.size(); ++j) {
         if (op->type == OpType_Input) {
@@ -74,13 +80,16 @@ void Program::createUnit(std::map<int, VARP>& varMap, std::vector<int>& inputInd
 //            int idx = outputIndexes[j];
             if (TensorDescribeName.find(newVar->name()) != TensorDescribeName.end()) {
                 int idx = TensorDescribeName[newVar->name()];
-                float scale = extraDescribes[idx]->quantInfo->scale;
-                float zero = extraDescribes[idx]->quantInfo->zero;
-                newVar->writeScaleMap(scale, zero);
+                if (idx >= 0 && idx < (int)extraDescribes.size() && extraDescribes[idx] && extraDescribes[idx]->quantInfo) {
+                    float scale = extraDescribes[idx]->quantInfo->scale;
+                    float zero = extraDescribes[idx]->quantInfo->zero;
+                    newVar->writeScaleMap(scale, zero);
+                }
             }
         }
         varMap[outputIndexes[j]] = newVar;
     }
+    return true;
 }
 
 VARPS Program::input(const std::unordered_map<std::string, VARP>& inputs, bool lazy) {
@@ -98,6 +107,10 @@ void Program::save(MNN::NetT* net) {
 }
 
 std::shared_ptr<Program> Program::create(const MNN::NetT* net, bool supportExtra, bool saveAllVars) {
+    if (nullptr == net) {
+        MNN_ERROR("Program::create: net is null\n");
+        return nullptr;
+    }
     return create(net->oplists, net->tensorName, net->outputName, supportExtra, saveAllVars, net);
 }
 
@@ -121,7 +134,10 @@ std::shared_ptr<Program> Program::create(const std::vector<std::unique_ptr<OpT>>
     }
     for (int index = 0; index < oplists.size(); ++index) {
         std::set<OpT*> invalidSet;
-        createUnit(varMap, inputIndexes, oplists, oplists[index].get(), tensorName, invalidSet, extraInputIndexes, net, TensorDescribeName);
+        if (!createUnit(varMap, inputIndexes, oplists, oplists[index].get(), tensorName, invalidSet, extraInputIndexes, net, TensorDescribeName)) {
+            MNN_ERROR("Program::create: failed to create op %s\n", oplists[index]->name.c_str());
+            return nullptr;
+        }
     }
     std::map<std::string, VARP> outputs;
     if (outputName.empty()) {
--- a/tools/converter/source/optimizer/Program.cpp
+++ b/tools/converter/source/optimizer/Program.cpp
@@ -32,8 +32,9 @@ bool Program::createUnit(std::map<int, VARP>& varMap, std::vector<int>& inputInd
     auto outputIndexes = op->outputIndexes;
     for (int j = 0; j < outputIndexes.size(); ++j) {
         if (varMap.find(outputIndexes[j]) != varMap.end()) {
-            // Don't support multi op output to one index
-            return false;
+            // Output may already exist when a dependency was created recursively while
+            // visiting an earlier op in oplists (graph is not necessarily topologically sorted).
+            return true;
         }
     }
     invalidSet.insert(op);
--- a/tools/converter/source/optimizer/Program.hpp
+++ b/tools/converter/source/optimizer/Program.hpp
@@ -27,7 +27,7 @@ public:
         return mOutputs;
     }
     VARPS input(const std::unordered_map<std::string, VARP>& inputs, bool lazy = false);
-    static void createUnit(std::map<int, VARP>& varMap, std::vector<int>& inputIndexes, const std::vector<std::unique_ptr<OpT>>& oplists, MNN::OpT* op, const MNN::NetT* net, std::set<OpT*>& invalidSet, std::set<int>& extraInputIndexes);
+    static bool createUnit(std::map<int, VARP>& varMap, std::vector<int>& inputIndexes, const std::vector<std::unique_ptr<OpT>>& oplists, MNN::OpT* op, const MNN::NetT* net, std::set<OpT*>& invalidSet, std::set<int>& extraInputIndexes);
 
     const std::map<int, VARP>& vars() const {
         return mVars;
@@ -36,7 +36,7 @@ public:
     void save(MNN::NetT* net);
 private:
     static std::shared_ptr<Program> create(const std::vector<std::unique_ptr<OpT>>& oplists, const std::vector<std::string>& tensorName, const std::vector<std::string>& outputName, bool supportExtra, bool saveAllVars, const MNN::NetT* net=nullptr);
-    static void createUnit(std::map<int, VARP>& varMap, std::vector<int>& inputIndexes, const std::vector<std::unique_ptr<OpT>>& oplists, MNN::OpT* op, const std::vector<std::string>& tensorName, std::set<OpT*>& invalidSet, std::set<int>& extraInputIndexes, const MNN::NetT* net=nullptr, std::map<std::string, int> TensorDescribeName = {});
+    static bool createUnit(std::map<int, VARP>& varMap, std::vector<int>& inputIndexes, const std::vector<std::unique_ptr<OpT>>& oplists, MNN::OpT* op, const std::vector<std::string>& tensorName, std::set<OpT*>& invalidSet, std::set<int>& extraInputIndexes, const MNN::NetT* net=nullptr, std::map<std::string, int> TensorDescribeName = {});
     Program() {
     }
     std::map<int, VARP> mVars;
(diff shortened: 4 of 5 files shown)


wangzhaode:
Thanks a lot for this PR, and for taking the time to fuzz the converter and file #4798 with a reproducer. The root cause you found is real and worth fixing. I built your branch locally (`-DMNN_BUILD_TEST=ON -DMNN_BUILD_CONVERTER=OFF`, macOS arm64) and the full unit suite passes (382/382), including the new `expr/create_guard` and all `expr/zeroshape*` cases, so the express-side guards look well behaved.

Before we can merge, though, there are a few things I'd like to ask you to change. The main concern is scope: this is filed as a converter hardening PR, but it also changes core runtime semantics, which makes it hard to review and hard to roll back independently.

### Blocking

**1. `MNNMemoryAllocAlign` / `MNNMemoryCallocAlign` zero-size behavior (`source/core/MNNMemoryUtils.cpp`)**

```cpp
if (0 == size || extra < alignment || size > SIZE_MAX - extra) { return NULL; }
```

The overflow guard is a good catch and we'd like to keep it. But `0 == size -> NULL` is a behavior change to a public allocator API: today `MNNMemoryAllocAlign(0)` returns a valid pointer, and callers that only check for `nullptr` will now treat a legitimate zero-size request as an allocation failure.

I checked the express call sites and they are all already guarded by `size > 0` (`express/Expr.cpp:176`, `express/Expr.cpp:770`), and `expr/zeroshape*` passes on your branch, so express itself is safe. But zero-count paths elsewhere (e.g. `ConvolutionCommon`) are not covered by the current tests, so the full blast radius isn't established.

Could you please keep only the overflow guard here and drop the `0 == size` condition? If you think the zero-size semantics should change, that's a reasonable discussion but it deserves its own PR with the rationale spelled out.

**2. Conflict with #4834 in `tools/converter/source/tflite/liteConverter.cpp`**

#4834 has just been merged. It touches the same lines as this PR (opcode index check, first output tensor index, the `inputs[k] < 0` check in the CUSTOM branch, the `delete op` paths) and it introduced a shared helper `tfliteAt<T>()` in `TfliteUtils.hpp`. Please rebase onto master and reuse `tfliteAt` instead of adding parallel checks.

There is also a philosophy difference worth aligning on: #4834 fails fast (`MNNNetT.reset(); return 1;`) whereas this PR tends to `continue` past bad entries. We'd prefer fail-fast — silently skipping a malformed entry produces a model that converts successfully but computes the wrong thing, which is much harder to diagnose than a hard conversion failure.

**3. `Program::createUnit` swallows failures (`tools/converter/source/optimizer/Program.cpp`)**

When the expr is null, `createUnit` returns early, which leaves holes in `varMap` while subsequent passes keep running on an incomplete graph. Please propagate the failure upward so that `Program::create` returns `nullptr` and the conversion aborts cleanly.

### Please double-check

**4. ONNX unresolved input is now a hard error (`tools/converter/source/onnx/onnxConverter.cpp`)**

Any non-empty input name that fails `scope->lookupTensor` now causes `return 1`. Are subgraph / outer-scope tensor references guaranteed to resolve through `lookupTensor`? If not this could reject models that convert fine today. Could you run the existing ONNX conversion regression set on this change before we take it?

### Suggested split

The `readModel()` fix is the highest-value part of this PR and I'd like to land it quickly on its own. The underlying problem — `LOG(FATAL)` not actually aborting, so `UnPackModel` runs on an unverified buffer — is exactly the root cause of #4798, and the accompanying `delete[] buffer`, empty-file guard and `oplists.empty()` check are all clean.

Would you be willing to split this into:

- **(a)** `readModel()` root-cause fix in `liteConverter.cpp`, rebased on the now-merged `tfliteAt` helper — we can review and merge this one fast;
- **(b)** `express/Expr.cpp` null-pointer guards plus the `expr/create_guard` test — small and self-contained;
- **(c)** `MNNMemoryUtils.cpp` overflow guard only (without the zero-size change);

and leave the ONNX change for last, once the regression set has been run?

Thanks again for the careful work here — the fuzzing effort is genuinely useful to us. Happy to review each piece as soon as it's up.

kabishou11:
Thanks for the detailed review and for merging #4834.

I'll follow the suggested split:

1. Rebase onto current master and land a focused `readModel()` / `liteConverter.cpp` fail-fast PR that reuses `tfliteAt` (no silent `continue` past malformed entries).
2. Follow with a small express null-guard + `expr/create_guard` PR.
3. Keep only the overflow guard in `MNNMemoryUtils.cpp` (drop the `0 == size → NULL` change).
4. Hold the ONNX unresolved-input hard error until I've run the ONNX conversion regression set.

I'll update this PR accordingly (or close it and open the smaller follow-ups) once the first piece is ready.

wangzhaode:
Thanks for narrowing the scope and addressing the earlier feedback! The split looks good.

One remaining issue is in `Program::createUnit()`: an output already being present in `varMap` does not necessarily mean the graph is invalid. For example, when A depends on B but appears before B in `oplists`, creating A recursively creates B. The outer loop later visits B with a fresh `invalidSet`, finds B's output already in `varMap`, and the new `return false` aborts conversion of this otherwise valid graph.

Could you please preserve the previous skip behavior here by returning `true` for this already-created case? Actual `Expr::create()` failures should still return `false` and propagate upward as this PR now does. Distinguishing genuinely conflicting producers can be handled separately if needed.

A small non-topologically ordered graph covering this revisit is enough to validate the fix; no broader test suite is needed. Thanks again!

kabishou11:
Thanks for the clarification — that makes sense.

I'll restore the previous skip behavior: if an output index is already in `varMap`, `createUnit` returns `true` (already created via a dependent's recursive create). Only a failed `Expr::create()` will return `false` and abort.

I'll also add a small non-topologically-ordered graph case covering the revisit path.

kabishou11:
@wangzhaode thank you again for the careful reviews on this PR (and on #4834).

I learned a lot from your feedback — not only the concrete `createUnit` / `varMap` revisit case, but also the broader way you think about converter hardening: keeping runtime semantics stable, preferring fail-fast over silent continues, and especially **splitting scope** so each change is reviewable and independently roll-backable. That framing is different from how I originally approached the patch, and it will stick with me on future converter work.

I'm applying the `return true` skip for already-created outputs now, with a small non-topo-ordered case as you suggested. Hope we get more chances to collaborate on MNN — your reviews have been genuinely helpful for sharpening both the code and the design judgment. Thanks again!

wangzhaode:
Thanks for fixing the revisit issue! The code change looks good now.

The small non-topological graph case was intended for local validation only; there is no need to commit a standalone test executable for it. Please remove `tools/converter/source/TestProgramCreate.cpp` and the corresponding `TestProgramCreate` target and link configuration from `tools/converter/CMakeLists.txt`. You can keep the local validation result in the PR discussion.

Sorry for not making that scope clear earlier. Once these test-only additions are removed, this is ready to merge. Thanks!

kabishou11:
Thanks for clarifying — removing `TestProgramCreate.cpp` and the CMake target now. Local validation already passed on my side (`TestProgramCreate` exited 0); I'll note that here so we don't need it in-tree.

Will push the cleanup shortly.

wangzhaode (approved):

PR #4839, opened 2026-09-09
[CPU:Perf] Enable RVV matrix kernels and optimize depthwise deconvolution

typer-J:
Register the RVV matrix and depthwise-deconvolution kernels in the `CoreFunctions` table under `MNN_USE_RVV`, and route `MNNMatrixProd` to its RVV kernel as well.

### How the kernels are installed

The RVV kernels carry `_RVV` names and are only written into the table inside the `if (gCoreFunction->supportRVV) { ... }` branch. The generic implementations in `ConvOpt.cpp` stay in place as the fallbacks, so a binary compiled with `MNN_USE_RVV=ON` still takes the scalar paths on a CPU without the V extension. This replaces the earlier revision of this branch, which removed the generic definitions and let the RVV objects take over the same symbols at link time.

### MNNMatrixProd was still dead code

Renaming `MNNMatrixProd` to `MNNMatrixProd_RVV` was not enough to make it live, because it has no `CoreFunctions` slot. `MNNMatrixAdd`, `MNNMatrixSub` and `MNNDeconvRunForUnitDepthWise` are reached through the table:

- `CPUDeconvolution.cpp:343` / `ConvInt8Winograd.cpp:253` → `core->MNNMatrixAdd`
- `StrassenMatmulComputor.cpp:144` → `core->MNNMatrixSub`
- `CPUDeconvolutionDepthwise.cpp:172` → `core->MNNDeconvRunForUnitDepthWise`

but `MNNMatrixProd` is a free function whose only caller, `MNNMatrixProdCommon`, called the bare generic symbol. The `_RVV` kernel therefore stayed unreachable. The dispatch now lives in `MNNMatrixProdCommon` and is guarded by a runtime `supportRVV` check — the same contract the table entries follow.

This is also why a direct-kernel test cannot prove the wiring: the previous revision's `MNNMatrixProd_RVV` cases passed while the kernel was never called. `RVVMatrixDeconvTest` now drives `MNNMatrixProdCommon` — the entry point used by `Matrix::prod`, the element-wise square in `CPUUnary` and `CPUMatrixBandPart` — against a scalar oracle, across widths that straddle the vector length and several row paddings.

Linkage check on the board (GCC 14.3.1, `-march=rv64gcv`, `-DMNN_USE_RVV`):

```
$ nm -C ConvOpt.o | grep MNNMatrixProd
T MNNMatrixProd
T MNNMatrixProdCommon
U MNNMatrixProd_RVV(float*, float const*, float const*, unsigned long, ...)
U MNN::MNNGetCoreFunctions()
```

### Depthwise deconvolution: the contiguous path needed a work threshold

`MNNDeconvRunForUnitDepthWise_RVV` has two paths and picks the contiguous one when `dilateX_step == 4` and `fw >= 16`. That path builds the repeated C4 pixel with an indexed load (`vloxei32`), which is a one-off setup of roughly 70ns on this board. With only the `fw >= 16` guard it could not amortise that setup: `fh = 1, fw = 16..32` measured 0.54x-0.87x.

The guard now also requires `fw * fh >= 48`, so those shapes stay on the unit-stride path, which is both faster and gather-free. Re-measured over the same 90 shapes, the deconvolution geomean went from 2.57x (min 0.54x, 6 regressions) to 2.90x (min 1.12x, none), and the `dilateX_step = 4, fh = 1` group improved from 1.43x to 2.75x.

### Validation

- Native GCC 14.3.1, VLEN=128, Release, rv64gcv, fast-math OFF, vendor IME2 OFF: MNN, `run_test.out`, `llm_demo` and `llm_bench` build successfully.
- Exact scalar/guard comparisons pass: 1 thread covers 1,368 matrix and 912 deconvolution cases; 4 threads cover 5,472 and 3,648 respectively. Includes aliases, zero sizes, tails, strides and overlapping rows. The same 2,280 single-thread cases pass under QEMU VLEN 128/256/512/1024.
- `MatMul`, `Deconvolution` and depthwise convolution operator tests pass with 1 and 4 threads. Final Qwen2.5-0.5B short/long fixed-history logits are byte-identical to the baseline.
- Kernel A/B against the actual generic implementations, re-run on the current head. Native GCC 14.3.1, VLEN=128, `-O3`. The generic side is compiled with the toolchain default march and the RVV side with `-march=rv64gcv`, mirroring the CMake split between the `MNNCPU` and `MNNRVV` targets -- building both with `-march=rv64gcv` lets GCC auto-vectorise the baseline and yields meaningless numbers.

  | kernel | geomean | min | shapes |
  |---|---|---|---|
  | `MNNMatrixAdd` | 3.53x | 2.94x | 12 |
  | `MNNMatrixSub` | 3.54x | 2.52x | 12 |
  | `MNNMatrixProd` | 3.57x | 2.49x | 12 |
  | `MNNDeconvRunForUnitDepthWise` | 2.90x | 1.12x | 90 |

  Deconvolution breaks down as 3.3x for `dilateX_step` 8 and 20, and 2.1x-2.8x for `dilateX_step` 4. No measured shape regresses. These are kernel measurements, not end-to-end model throughput.
- Qwen2.5-0.5B short/long fixed-history logits remain byte-identical to the baseline, so the deconvolution threshold change does not move model output.

Wider native VLEN and fast-math are not validated here.

Reproduction (`test/backend/cpu/RVVMatrixDeconvTest.cpp`, `-DMNN_RVV_MATRIX_TEST_MAIN`):

```sh
$CXX -std=c++11 -O3 -march=rv64gcv -ffp-contract=off -pthread \
 -Iinclude -Isource -DMNN_RVV_MATRIX_TEST_MAIN \
 source/backend/cpu/riscv/rvv/MNNMatrixAdd.cpp \
 source/backend/cpu/riscv/rvv/MNNMatrixSub.cpp \
 source/backend/cpu/riscv/rvv/MNNMatrixProd.cpp \
 source/backend/cpu/riscv/rvv/MNNDeconvRunForUnitDepthWise.cpp \
 test/backend/cpu/RVVMatrixDeconvTest.cpp -o rvv-matrix-test
./rvv-matrix-test 128 1
./rvv-matrix-test 128 4
```

The first argument must match the measured VLEN; repeat at 256/512/1024 under QEMU.

```
```


wangzhaode:
Thanks for the contribution! The matrix and depthwise-deconvolution kernel checks passed on K3 (VLEN=256), including the 1/4-thread checks.

One runtime fallback issue needs fixing before merging: MNN_USE_RVV currently removes the generic MatrixAdd/MatrixSub/MatrixProd and depthwise-deconvolution implementations and replaces the same symbols with RVV code. The default function table therefore already points to RVV implementations even when the runtime supportRVV check is false.

Please keep the generic implementations, give the RVV variants distinct names, and override the corresponding function-table entries only inside the supportRVV branch. This preserves the scalar path for an RVV-enabled binary running on a CPU without RVV. A focused dispatch/fallback check is enough. Thanks!

typer-J:
Fixed — the runtime fallback now works exactly as you described. Latest commit: 6525d4d96b.

Changes:

- `ConvOpt.cpp`: removed all four `#ifndef MNN_USE_RVV` guards, so the generic `MNNMatrixAdd`, `MNNMatrixSub`, `MNNMatrixProd` and `MNNDeconvRunForUnitDepthWise` are present unconditionally again.
- The RVV kernels now carry distinct names: `MNNMatrixAdd_RVV`, `MNNMatrixSub_RVV`, `MNNMatrixProd_RVV`, `MNNDeconvRunForUnitDepthWise_RVV`. No symbol is replaced at link time anymore.
- In `CommonOptFunction.cpp`, the three table entries are overridden only inside the `if (gCoreFunction->supportRVV)` branch:
  `MNNMatrixAdd`, `MNNMatrixSub` and `MNNDeconvRunForUnitDepthWise`.
  A binary built with `MNN_USE_RVV=ON` running on a CPU where `supportRVV` is false therefore keeps the scalar implementations.

One detail worth flagging: `MNNMatrixProd` is not a `CoreFunctions` member — it is only called by `MNNMatrixProdCommon`, a free function in the same TU. With the generic definition restored and the RVV copy renamed, that internal call now resolves to the scalar implementation, which is the safe outcome. Making prod itself table-dispatched would mean threading `CoreFunctions*` into `MNNMatrixProdCommon` and updating its three call sites (`CPUMatrixBandPart`, `CPUUnary`, `Matrix.cpp`) — happy to do it if you would rather have the vector path for prod, but it seemed outside the scope of the fallback fix.

I also had to update `test/backend/cpu/RVVMatrixDeconvTest.cpp`: it declared the old generic symbols via `extern "C"`, so after the rename it still compiled and passed but was quietly exercising the scalar path. It now points at the `_RVV` kernels.

Rebase note: this PR touches the same registration block in `CommonOptFunction.cpp` as the other RVV PRs, so I expect a conflict there — will rebase as they land.

typer-J:
## 板端复验：修正一处链接错误（`extern` 声明命名空间位置）

在 `riscv64` 目标板上做原生构建时，`rvv` 变体链接失败（`off` 变体正常）：

```
ld: CommonOptFunction.cpp:(.text._ZN3MNN19MNNCoreFunctionInitEv+0x632):
    undefined reference to `MNN::MNNMatrixAdd_RVV(float*, float const*, ...)'
ld: ... `MNN::MNNMatrixSub_RVV(...)'
ld: ... `MNN::MNNDeconvRunForUnitDepthWise_RVV(...)'
```

**根因**：链接器在找带命名空间前缀的 `MNN::MNNMatrixAdd_RVV`，但这三个内核是定义在
**全局命名空间**里的，我把 `extern` 声明放进了 `namespace MNN {` 之内，导致名字修饰不一致。

`nm` 实证：

| | 符号 |
|---|---|
| 内核 `.o` 实际导出 | `_Z16MNNMatrixAdd_RVVPfPKfS1_mmmmm`（**无** `N3MNN`） |
| 链接器索要 | `_ZN3MNN16MNNMatrixAdd_RVV...`（**有** `N3MNN`） |

对照本仓库里已有的正确写法（`#4821` 的 `extern void MNNAbsMaxFP32_RVV(...)` 就位于
`namespace MNN {` 之外），已把这三条声明移到 `namespace MNN {` 之前：

```cpp
extern void MNNGetMatMulPackMode_RVV(int* eP, int* lP, int* hP);
// RVV kernels below live in the global namespace, so their declarations must
// stay outside namespace MNN as well.
extern void MNNMatrixAdd_RVV(float* C, const float* A, const float* B, size_t widthC4, size_t cStride,
                             size_t aStride, size_t bStride, size_t height);
extern void MNNMatrixSub_RVV(...);
extern void MNNDeconvRunForUnitDepthWise_RVV(...);
namespace MNN {
void MNNRvvInitializeFastPathFunctions(CoreFunctions* core);
}
```

板端复验结果：

| 变体 | 构建 | 测试（`op/Deconvolution` / `op/convolution/depthwise_conv` 等） |
|---|---|---|
| `MNN_USE_RVV=ON` | ✅ `exit=0`（此前链接失败） | ✅ 全通过 |
| `MNN_USE_RVV=OFF` | ✅ `exit=0` | ✅ 全通过 |

```
$ ./run_test.out op/Deconvolution 0 1 1
op/Deconvolution cost time: 0.136 ms
op/DeconvolutionInt8 cost time: 0.738 ms
√√√ all <op/Deconvolution> tests passed.
TEST_CASE_AMOUNT_UNIT: {"blocked":0,"failed":0,"passed":3,"skipped":0}
```

环境：riscv64 / openEuler 24.03 LTS-SP3 / 32 核 / GCC 14.3.1 / VLEN=128。

typer-J:
#4839 现在按 #4847 的模式落地：

- `test/backend/cpu/RVVMatrixDeconvTest.cpp` → 改为 `MNNTestRVVMatrixDeconvFunctions()`，由 `test/op/DeconvolutionTest.cpp` 的 `op/Deconvolution` 调用，移除未定义的 `MNN_RVV_MATRIX_TEST_MAIN` 守卫与独立 `main()`。注意 `DECL_MATRIX` 用 `name##_RVV` 拼接，而不是字面量 `name_RVV`。
- `test/op/DeconvolutionTest.cpp` 的 `DeconvolutionTest::run()` 末尾返回该 helper（`op/Deconvolution` 已注册）。
- `test/CMakeLists.txt` 加 `MNN_TEST_RVV_ENABLED=$<BOOL:${MNN_USE_RVV}>` 编译定义。

- `pr4839-rvv / off × op/Deconvolution t=1,4`：rvv 变体看到 `RVV matrix/Deconv: matrix_cases=5472 deconv_cases=3648 RVV=1`，注册断言 `core->MNNMatrixAdd != MNNMatrixAdd_RVV` 等三条全过；off 变体看到 `RVV matrix/Deconv: skipped, MNN_USE_RVV=OFF (RVV=1)`。

注册断言在 `supportRVV=1` 时执行；`supportRVV=0` 时跳过，避免误报。

新增的 commit：
- `[RVV:Test] Move matrix and deconv kernel test into run_test.out`
- `[RVV:Test] op/Deconvolution runs the matrix/Deconv kernel test`

请重新 review。如仍不合入，请直接指出还要改哪一处。

typer-J:
Thank you for your guidance and careful review. I have fixed the linkage issue and integrated the RVV regression checks into the existing test suite.

The RVV-enabled build initially failed to link because the declarations for `MNNMatrixAdd_RVV`, `MNNMatrixSub_RVV`, and `MNNDeconvRunForUnitDepthWise_RVV` were placed inside `namespace MNN`, while the kernel definitions are in the global namespace. The declarations have been moved outside `namespace MNN`, so their C++ symbol names now match the definitions.

For the tests:

- `RVVMatrixDeconvTest.cpp` now exposes `MNNTestRVVMatrixDeconvFunctions()` instead of relying on the undefined `MNN_RVV_MATRIX_TEST_MAIN` guard and a standalone `main()`.
- The existing registered `op/Deconvolution` test calls the helper, so the checks run through the normal `run_test.out` path.
- `test/CMakeLists.txt` defines `MNN_TEST_RVV_ENABLED` from `MNN_USE_RVV`.
- Dispatch assertions cover the RVV replacements that are actually registered in `CoreFunctions`: MatrixAdd, MatrixSub, and depthwise Deconvolution. MatrixProd is tested directly but is not asserted as a table entry because it is not a `CoreFunctions` member.

Native riscv64 verification (openEuler 24.03, GCC 14.3.1, VLEN=128) completed successfully:

- Both `MNN_USE_RVV=ON` and `OFF` builds compiled and linked successfully.
- `op/Deconvolution` passed with 1 and 4 test threads in both variants.
- The RVV build executed 5,472 matrix cases and 3,648 Deconvolution cases.
- The OFF build reached the same registered test and correctly skipped the RVV-only checks.

I appreciate your feedback on preserving the scalar fallback and validating the real dispatch path. Please let me know if any further adjustment is needed.

typer-J:
Thank you for the detailed fallback requirement. Here is a summary of what was changed in response to your review, and the results.

**Changes (latest commit `a0fd5ff3`):**

- The generic `MNNMatrixAdd`, `MNNMatrixSub`, `MNNMatrixProd` and `MNNDeconvRunForUnitDepthWise` implementations are present unconditionally again (the `#ifndef MNN_USE_RVV` guards were removed), so no symbol is replaced at link time.
- The RVV kernels carry distinct names (`MNNMatrixAdd_RVV`, `MNNMatrixSub_RVV`, `MNNMatrixProd_RVV`, `MNNDeconvRunForUnitDepthWise_RVV`), and their declarations were fixed to match the definition linkage so RVV-enabled builds link correctly.
- The function-table entries for MatrixAdd, MatrixSub and depthwise Deconvolution are overridden only inside the `supportRVV` branch, preserving the scalar path for an RVV-enabled binary running on a CPU without RVV. (`MNNMatrixProd` is not a `CoreFunctions` member; its internal call now resolves to the scalar implementation, which is the safe outcome.)
- The matrix/deconvolution checks were moved into the existing registered `op/Deconvolution` test (via `MNNTestRVVMatrixDeconvFunctions()` and `MNN_TEST_RVV_ENABLED`), with dispatch assertions covering the registered table entries.

**Results:**

- On a riscv64 board (openEuler 24.03, GCC 14.3.1, VLEN=128): both `MNN_USE_RVV=ON` and `OFF` builds compiled and linked successfully, and `op/Deconvolution` passed with 1 and 4 threads in both variants. The RVV build executed 5,472 matrix cases and 3,648 Deconvolution cases with the dispatch assertions confirming `RVV=1`; the OFF build reached the same registered test and correctly skipped the RVV-only checks.

Thank you again for the review. Please let me know if anything else needs adjusting.

typer-J:

We have opened #4864 with a minimal one-line fix (exporting `MNNGetCoreFunctions()` via the existing `MNN_PUBLIC` macro). With that fix applied, we verified on a riscv64 board that the shared-library + RVV configuration builds successfully and this PR's tests (`op/Deconvolution`, 1 and 4 threads) pass, including the matrix/deconvolution dispatch assertions.

typer-J:
Thank you for the precise fallback requirement. Both points from the review are addressed, and measuring the result turned up one more problem we would otherwise have shipped.

## 1. Generic implementations were being replaced (your main point)

You were right: `MNN_USE_RVV` removed the generic `MNNMatrixAdd` / `MNNMatrixSub` / `MNNMatrixProd` / `MNNDeconvRunForUnitDepthWise` implementations and replaced the same symbols with RVV code, so the default function table already pointed at RVV even when the runtime `supportRVV` check is false.

- The generic implementations are present unconditionally again (the `#ifndef MNN_USE_RVV` guards were removed), so nothing is replaced at link time.
- The RVV kernels carry distinct names (`*_RVV`) and their declarations were fixed to match the definition linkage.
- The table entries for MatrixAdd, MatrixSub and depthwise deconvolution are overridden only inside the `supportRVV` branch, preserving the scalar path for an RVV-enabled binary on a CPU without RVV.
- `MNNMatrixProd` is not a `CoreFunctions` member, so renaming alone was not enough — `MNNMatrixProdCommon` now dispatches on `supportRVV` and falls back to the generic function. That was the one case where the rename left dead code behind, and it is now exercised by the test.
- The checks moved into the registered `op/Deconvolution` test with dispatch assertions covering the table entries.

## 2. Performance (kernel A/B on the current head, native RVV VLEN=128, GCC 14.3.1)

The generic side is built with the toolchain default march and the RVV side with `-march=rv64gcv`, mirroring the CMake split between the `MNNCPU` and `MNNRVV` targets — building both with `-march=rv64gcv` lets GCC auto-vectorise the baseline and yields meaningless numbers.

| kernel | geomean | min | shapes |
|---|---:|---:|---:|
| `MNNMatrixAdd` | 3.53x | 2.94x | 12 |
| `MNNMatrixSub` | 3.54x | 2.52x | 12 |
| `MNNMatrixProd` | 3.57x | 2.49x | 12 |
| `MNNDeconvRunForUnitDepthWise` | 2.90x | 1.12x | 90 |

The deconvolution figure needed a fix that only showed up under measurement. The contiguous path (`dilateX_step == 4`) builds the repeated C4 pixel with an indexed load, a one-off setup of roughly 70 ns on this board, so with only the `fw >= 16` guard it could not amortise that setup and lost it back at `fh = 1, fw = 16..32` (0.54x-0.87x). The guard now also requires `fw * fh >= 48`, so those shapes stay on the unit-stride path. Over the same 90 shapes the geomean went from 2.57x (min 0.54x, 6 regressions) to 2.90x (min 1.12x, none).

Correctness: `op/Deconvolution` passes with 1 and 4 threads in both `MNN_USE_RVV=ON` and `OFF` builds; the RVV build executed 5,472 matrix cases and 3,648 deconvolution cases with the dispatch assertion confirming `RVV=1`, and Qwen2.5-0.5B fixed-history logits remain byte-identical to the baseline. `MNNMatrixProd` is covered by the registered test for the first time, through `MNNMatrixProdCommon`.

## 3. Build configuration note

A default shared-library build with `MNN_USE_RVV=ON` fails to link `run_test.out` on current master (`undefined reference to MNN::MNNGetCoreFunctions()`), a regression from our already-merged #4841/#4842. The one-line `MNN_PUBLIC` export is now carried in this PR itself rather than in a separate PR.

Thank you again — the distinct-name plus runtime-guard contract is exactly what made the `MNNMatrixProd` dead-code case visible.

wangzhaode (approved):

PR #4847, opened 2026-09-10
[CPU:Perf] Vectorize unpacked E=1 GEMV on RISC-V

typer-J:
CPUMatMul already bypasses packing for E=1, but its non-transposed B path still uses the generic scalar GEMV. Add an RVV kernel that vectorizes independent output channels in raw B[K,H], assigns disjoint vector blocks to workers, and registers it behind the existing RVV build/runtime gates. The input/output ABI and Tensor packing stay unchanged; transposed B delegates to the original function.

Native validation on a RISC-V server with GCC 14.3.1 and measured VLEN=128:

- Fresh static MNN builds with RVV ON/OFF, LLM, LOW_MEMORY and TRANSFORMER_FUSE enabled; IME2 and RVV fast-math disabled.
- Eight MatMul test runs across ON/OFF, 1/4 threads and precision settings 1/2: 16 tests passed, including matmulBConst. Each run additionally verifies exact function-table registration and 2,048 thread/tail/special-value configurations. Unchanged transposed and RVV-OFF paths are checked bitwise.
- Same-process original-scalar/production-RVV comparison, three fresh processes per default/strict floating-point configuration. For the 32 non-transposed H>1 configurations that use the new dispatcher path, default compilation gives a 1.864x geometric-mean kernel speedup; 31 improve by more than 5% in every process, with no stable regression above 5%. This is single-thread kernel evidence, not an end-to-end model result.
- The new production function is 170 bytes in the measured object and introduces no dynamic global initialization or new global test registration object.

The kernel uses separate multiply/add intrinsics. Target-result checks use the existing MatMul numerical-error contract: default scalar FMA can differ in the sign of an underflowed zero, demonstrated by a retained minimal reproducer. Special inputs remain covered; no blanket bitwise-equivalence claim is made for the new arithmetic path.

Reproduction: `test/backend/cpu/RVVMatMulTest.cpp`, registered with the normal test suite and run through `run_test.out` on an RVV build (`MNN_BUILD_STATIC_LIBS` plus `__riscv`). The standalone benchmark harness and its report are not part of this PR.

```
```


wangzhaode:
Thanks for the contribution! We validated this on K3 (VLEN=256): the E=1 GEMV dispatch tests passed with 1 and 4 threads, and the non-transposed kernel achieved about 1.86x geometric-mean speedup in the microbenchmark.

Before merging, please remove the entire newly added benchmark/rvv_gemv/ directory from this PR, including the scripts, sources, README, report and .gitignore. The validation results can stay in the PR discussion; please keep the production changes and the existing test-suite coverage. Please also rebase if needed after the other RVV PRs are merged. Thanks!

typer-J:
Done — `benchmark/rvv_gemv/` is removed in 2d223dc673.

The commit drops the whole directory (`.gitignore`, `README.md`, `report`, `run.py`, `benchmark.cpp` and the related sources) and nothing else: the production kernel changes and the existing test-suite coverage are untouched. Diff is a pure deletion of the benchmark tooling, -359 lines across 7 files.

The validation numbers you measured can stay here in the discussion as you suggested. Happy to rebase once the other RVV PRs land.

typer-J:
Thank you for the review. Here is a summary of what was changed in response to your comments, and the results.

**Changes (latest commit `a1acade9`):**

- The whole `benchmark/rvv_gemv/` directory was removed as requested (scripts, sources, README, report and `.gitignore`), with the production kernel changes and the existing test-suite coverage untouched. The validation results remain in the PR description and discussion.
- The E=1 GEMV checks are reachable through the existing registered `op/matmul` test (via `MNNTestRVVMatMulFunctions()` and `MNN_TEST_RVV_ENABLED`), including an exact function-table dispatch assertion.

**Results:**

- On a riscv64 board (openEuler 24.03, GCC 14.3.1, VLEN=128): both `MNN_USE_RVV=ON` and `OFF` builds compiled and linked successfully, and `op/matmul` passed with 1 and 4 threads in both variants. The RVV build's dispatch assertion confirmed the table points at the RVV kernel (2,048 configurations passed, `RVV=1`).

**One upstream observation (independent of this PR):**

While verifying all five RVV PRs against master `a03b005c`, we found that a default shared-library build with `MNN_USE_RVV=ON` fails to link `run_test.out` on current master, with `undefined reference to MNN::MNNGetCoreFunctions()`. The references come from master's own `test/backend/cpu/RVVAttentionMaskTest.cpp` and `RVVLinearAttentionTest.cpp`, which are activated because the current `test/CMakeLists.txt` defines `MNN_USE_RVV` on `run_test.out` when the `MNNRVV` target exists. Since `libMNN.so` is built with `-fvisibility=hidden` and this internal symbol has no default-visibility annotation, it is not exported (`nm -D` confirms), so the shared build cannot resolve it. A pure-master tree (no PR changes) reproduces the same failure, while static builds are unaffected. It may be worth exporting the symbol or gating those tests to avoid breaking shared-library RVV builds. Happy to help verify any fix.

Thank you again for the review. Please let me know if anything else needs adjusting.

typer-J:
A follow-up on the upstream observation above: we have opened #4864 with a minimal one-line fix — exporting `MNNGetCoreFunctions()` via the existing `MNN_PUBLIC` macro, so the shared library actually exports the symbol that the new RVV tests reference.

With that fix applied, we verified on a riscv64 board (openEuler 24.03, GCC 14.3.1, VLEN=128) that the previously failing shared-library + `MNN_USE_RVV=ON` configuration builds successfully, and the affected tests (`backend/cpu/rvv/attention_mask`, `backend/cpu/rvv/linear_attention`) as well as this PR's `op/matmul` all pass with 1 and 4 threads.

May we kindly suggest merging #4864 first? Once it lands, this PR's RVV-enabled shared-library variant becomes buildable as configured by the current `test/CMakeLists.txt`, and the remaining RVV PRs benefit in the same way. Apologies again for the regression introduced by #4841/#4842, and thank you for your time.

typer-J:
Thank you for the review and for validating this on K3.

## 1. Your request — done

The entire `benchmark/rvv_gemv/` directory is removed from this PR: scripts, sources, README, report and `.gitignore`. The production kernel changes and the existing test-suite coverage are untouched, and the validation results stay in the PR description and this discussion, as you suggested. The PR was rebased onto master as a single commit.

For the same reason, the standalone notes that were attached to the other RVV PRs have been removed as well; the reproduction commands now live in each PR description instead of in separate files.

## 2. Call path

The E=1 GEMV checks are reachable through the existing registered `op/matmul` test (`MNNTestRVVMatMulFunctions()` under `MNN_TEST_RVV_ENABLED`), including an exact function-table dispatch assertion, so they run through the normal `run_test.out` path in both build variants rather than as a separate harness.

## 3. Performance (kernel level, native RVV VLEN=128, GCC 14.3.1, three fresh processes per default / strict floating-point configuration)

For the 32 non-transposed `H>1` configurations that use the new dispatcher path, default compilation gives a 1.864x geometric-mean kernel speedup; 31 of the 32 improve by more than 5% in every process, with no stable regression above 5%. This matches the order of magnitude you measured on K3. These are single-thread kernel results, not an end-to-end model claim.

`MNN_USE_RVV=ON` and `OFF` builds both compile and link, and `op/matmul` passes with 1 and 4 threads in both; the RVV build's dispatch assertion confirmed the table points at the RVV kernel (2,048 configurations passed, `RVV=1`).

While verifying all the RVV PRs against master `a03b005c`, we found that a default shared-library build with `MNN_USE_RVV=ON` fails to link `run_test.out` with `undefined reference to MNN::MNNGetCoreFunctions()`. The references come from master's own `test/backend/cpu/RVVAttentionMaskTest.cpp` and `RVVLinearAttentionTest.cpp`, which are activated because `test/CMakeLists.txt` defines `MNN_USE_RVV` on `run_test.out` when the `MNNRVV` target exists. `libMNN.so` is built with `-fvisibility=hidden` and this internal symbol has no default-visibility annotation, so it is not exported (`nm -D` confirms); a pure-master tree reproduces the failure and static builds are unaffected.

Rather than keeping that as a separate PR, the one-line `MNN_PUBLIC` export is now carried directly in this PR, so a shared-library + `MNN_USE_RVV=ON` build works as soon as this lands. Apologies again for the regression our earlier merged PRs introduced.

Thank you for your time.

wangzhaode (approved):

PR #4848, opened 2026-09-10
[LLM:Refactor] Remove the unused duplicate of OnnxRebuilder

ErenAta16:
## What this is

`transformers/llm/export/utils/onnx.py` and
`transformers/llm/export/utils/onnx_rebuilder.py` are the same file. Not similar,
identical:

```
56aa54e6f20156fd5128ce77391d4227  utils/onnx.py
56aa54e6f20156fd5128ce77391d4227  utils/onnx_rebuilder.py
```

Both are 69 lines and both define `OnnxRebuilder`. Only one of them is wired up:

```
llmexport.py:18   from utils.onnx_rebuilder import OnnxRebuilder
llmexport.py:523  return OnnxRebuilder(onnx_path, self.unloaded_ops).rebuild()
```

Grepping the repository for `utils.onnx`, `from utils import onnx`, `from .onnx
import` and dynamic forms turns up nothing, and `utils/__init__.py` registers
neither module. So `utils/onnx.py` has no importer.

Both files entered in `76681528`, "MNN:Sync: Sync Internal 3.0.4" on 2025-01-22,
which suggests the copy came along with the sync rather than being added on
purpose.

This PR deletes the unimported one and changes nothing else.

## Why it is worth doing

Mainly that a second copy of a class is a place for the two to drift. A fix
applied to `onnx_rebuilder.py` would leave `onnx.py` behind, and since the dead
one is byte-identical today, a reader has no way to tell from the file which is
the live one.

There is also a smaller latent thing. `utils/onnx.py` does `import onnx` on line 2,
so the module shadows the package it imports if `utils/` ever lands on `sys.path`
directly, at which point that import resolves to the file itself. Nothing does
that today, `gguf/gguf_reader.py:26` inserts the `export/` directory rather than
`utils/`, so this is a hazard rather than a live bug and I am not claiming
otherwise.

## Verification

Checked on a clone at `main`. After deleting the file,
`python -m py_compile` is clean on both `utils/onnx_rebuilder.py` and
`llmexport.py`, and no reference to `utils.onnx` remains anywhere in the tree.

I have not run the full export pipeline, which needs model weights I do not have
here, so treat that as unverified. The claim this PR rests on is narrower: the
deleted file had no importer, which is checkable by grep.


ErenAta16:
Closing this, since I am not in a position to complete the CLA step and it would
be worse to leave a PR parked against a gate it cannot pass.

The finding does not depend on the PR, so leaving it here in case it is useful to
whoever touches that directory next:

```
56aa54e6f20156fd5128ce77391d4227  transformers/llm/export/utils/onnx.py
56aa54e6f20156fd5128ce77391d4227  transformers/llm/export/utils/onnx_rebuilder.py
```

The two files are byte-identical and both define `OnnxRebuilder`. Only
`utils.onnx_rebuilder` is imported, at `llmexport.py:18`, and used at
`llmexport.py:523`. Nothing in the tree references `utils.onnx` and
`utils/__init__.py` registers neither. Both arrived together in `76681528`
("Sync Internal 3.0.4", 2025-01-22), so the copy looks like an artifact of the
sync rather than a deliberate addition.

Deleting `utils/onnx.py` is the whole change. After it, `python -m py_compile` is
clean on `utils/onnx_rebuilder.py` and `llmexport.py`, and no reference to
`utils.onnx` remains. That is a one-line commit for anyone on the team, no PR
needed.

Thanks for the `CONTRIBUTING.md` section on AI-assisted contributions, by the way.
Repos that say plainly whether that is welcome are easier to contribute to than
ones that leave it to be guessed at, and it is not common to see it stated.

PR #4863, opened 2026-09-11
[Pymnn:Test] Remove the dead cv.calcHist copy of test_histogram

Anai-Guo:
## Problem

`UnitTest` in `pymnn/test/unit_test.py` defines `test_histogram` **twice** inside the same class body:

| line | section | assertion |
|---|---|---|
| 804 | `MNN.cv` | `cv.calcHist([self.img], [0], None, [257], [0., 256.])` == `cv2.calcHist(...)` |
| 1205 | `MNN.numpy` | `mp.histogram(x, 7, (2, 4))` == `np.histogram(...)` |

Python binds class-body names in order, so the second `def` replaces the first and only the `mp.histogram` case is ever collected:

```
$ python -m pyflakes pymnn/test/unit_test.py
pymnn/test/unit_test.py:1205:1: redefinition of unused 'test_histogram' from line 804

def statements named test_* : 181
distinct names              : 180
unittest collects           : 180
```

## Why it is deleted rather than renamed

The obvious repair would be to rename the shadowed copy (as [`42e3160d`](https://github.com/alibaba/MNN/commit/42e3160d3dce879493cf2db8ad6f305ebd4705cd) *"Fix duplicate test methods"* did for `test_normalize`, `test_resize` and `test_split` in this same file). That does not work here, because the shadowed copy's **expectation is wrong**.

`CV::calcHist` forwards to `_Histogram` and is intentionally not bin-compatible with OpenCV — `tools/cv/source/imgproc/histograms.cpp` says so in the implementation itself:

```cpp
// last bin deal different with opencv, same as torch and numpy
return _Histogram(images[0], histSize[0], int(ranges[0]), int(ranges[1]), channels[0]);
```

Measured against the released wheel (`MNN==3.6.1`) on `resource/images/cat.jpg`, with exactly the arguments the shadowed test uses:

```
mnn shape (257,)   cv2 shape (257,)
max abs diff        10.0
bins differing      193 / 257
total count         172800  (both)
first 12 mnn        [  9.   5.  13.  25.  33.  60.  89. 118. 137. 196. 239. 293.]
first 12 cv2        [  9.   5.  13.  25.  33.  60.  89. 118. 135. 197. 240. 293.]
```

Equal totals, different distribution — a binning difference, which is precisely the documented divergence. So `assertEqualVar(cv.calcHist(...), cv2.calcHist(...))` has never held; it survives in the tree only because it is unreachable. Renaming it would turn `pymnn-linux` / `pymnn-macos` / `pymnn-windows` red, since those all run `python3 unit_test.py` on any PR touching `pymnn/**`.

Deleting it removes a case that asserts behaviour MNN deliberately does not implement, and removes the trap of a test body that looks maintained but cannot run.

## Change

```diff
-    # histogram
-    def test_histogram(self):
-        hist = cv.calcHist([self.img], [0], None, [257], [0., 256.])
-        hist_ = cv2.calcHist([self.img_], [0], None, [257], [0., 256.])
-        self.assertEqualVar(hist, hist_)
     # calib3d
```

`MNN.numpy.histogram` keeps `test_histogram` unchanged and remains the coverage for the numpy-compatible path. Collection stays at 180 — nothing that runs today is affected.

If you would rather keep a `MNN.cv.calcHist` case, I am happy to replace this with one that asserts the numpy/torch binning that `calcHist` actually implements (i.e. compares against `np.histogram` instead of `cv2.calcHist`) — just say which you prefer.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)


wangzhaode:
Thanks for the cleanup and the clear explanation! Keeping the MNN.numpy.histogram test and removing the shadowed cv.calcHist test looks good to us; no further code changes are needed.

The CLA check is still pending. Could you please sign the [Contributor License Agreement](https://cla-assistant.io/alibaba/MNN?pullRequest=4863)? If you have already signed it, please use the recheck link in the CLA bot comment. Once the check passes, we can proceed with merging. Thanks again!

PR #4864, opened 2026-09-12
[CPU:Bugfix] Export MNNGetCoreFunctions for shared-library RVV test builds

typer-J:
## Summary

This fixes a shared-library build regression that our merged PRs #4841 and #4842 introduced, for which we apologize. The regression breaks `run_test.out` linking when MNN is built as a shared library with `MNN_USE_RVV=ON`.

## Root cause

PRs #4842 and #4841 added `test/backend/cpu/RVVLinearAttentionTest.cpp` and `RVVAttentionMaskTest.cpp`, and #4842 also changed `test/CMakeLists.txt` to define `MNN_USE_RVV` on `run_test.out` when the `MNNRVV` target exists. With that macro active, both tests reference `MNN::MNNGetCoreFunctions()`. However, `libMNN.so` is compiled with `-fvisibility=hidden` and this symbol has no visibility annotation, so it is not exported from the shared library (`nm -D` confirms). The link therefore fails with `undefined reference to MNN::MNNGetCoreFunctions()` on any default shared-library build with RVV enabled. Static builds are unaffected, which is why this was not caught earlier. A pure master tree reproduces the failure independently of any other PR.

## Fix

A minimal one-line change: annotate the `MNNGetCoreFunctions()` declaration in `source/backend/cpu/compute/CommonOptFunction.h` with the existing `MNN_PUBLIC` macro (`__attribute__((visibility("default")))` on non-MSVC), so the symbol is exported from the shared library and the RVV tests can resolve it. No behavior change for static builds or for the library's internal callers.

## Verification

On a riscv64 board (openEuler 24.03, GCC 14.3.1, VLEN=128):

- Pure master `a03b005c`, shared library + `MNN_USE_RVV=ON`: `run_test.out` fails to link with `undefined reference to MNN::MNNGetCoreFunctions()` (reproduces the regression).
- With this fix applied, the same configuration builds successfully, and `backend/cpu/rvv/attention_mask` and `backend/cpu/rvv/linear_attention` tests run through the normal `run_test.out` path.
- Static builds with `MNN_USE_RVV=ON`/`OFF` remain unaffected.

## Module

CPU/Core

## Type

- [x] Bugfix

### DIFF
--- a/source/backend/cpu/compute/CommonOptFunction.h
+++ b/source/backend/cpu/compute/CommonOptFunction.h
@@ -588,7 +588,7 @@ struct CoreFunctions {
     const CPUExtension* extension = nullptr;
 };
 void MNNCoreFunctionInit();
-CoreFunctions* MNNGetCoreFunctions();
+MNN_PUBLIC CoreFunctions* MNNGetCoreFunctions();
 }; // namespace MNN
 
 #endif /* CommonOptFunction_h */


typer-J:
Hello @wangzhaode,

We would like to sincerely apologize for the trouble caused by our merged PRs #4841 and #4842. When adding the RVV attention/linear-attention kernels, we referenced `MNN::MNNGetCoreFunctions()` from the new tests and enabled them via a `MNN_USE_RVV` definition on `run_test.out`, without realizing that this symbol is not exported from the shared library under `-fvisibility=hidden`. Our own validation used static builds, so the shared-library link failure slipped through both our testing and review — that was our oversight, and we are sorry for the inconvenience it may cause anyone building MNN as a shared library with RVV enabled.

This PR fixes the regression with a minimal one-line change (exporting `MNNGetCoreFunctions()` via the existing `MNN_PUBLIC` macro). It has been verified on a riscv64 board: the previously failing shared-library + `MNN_USE_RVV=ON` configuration now builds and the affected tests pass with 1 and 4 threads, while static builds remain unaffected.

We have also double-checked our other open RVV PRs — their tests follow the same function-table pattern, and with this fix in place all of them build and pass in both shared and static configurations. Thank you for your patience and guidance, and again, our apologies for the regression.

typer-J:
Folded into #4821, #4838, #4839, #4846, #4847 and #4865 -- each of those branches now carries the same one-line change (`MNN_PUBLIC` on `MNNGetCoreFunctions()` in `source/backend/cpu/compute/CommonOptFunction.h`). That makes every one of them self-contained and independently mergeable, so closing this standalone PR rather than merging the identical hunk in seven places.

Two notes:

- The regression this fixes is already in master: the merged #4841/#4842 tests (`RVVAttentionMaskTest.cpp`, `RVVLinearAttentionTest.cpp`) use the same `RVV_TEST_KERNEL` macro, and `test/CMakeLists.txt` defines `MNN_USE_RVV` on `run_test.out`. So master stays broken for shared-library + `MNN_USE_RVV=ON` builds until the first of those PRs lands.
- Static builds are unaffected, which is why this was not caught earlier.

PR #4865, opened 2026-09-12
[CPU:Perf] Add an RVV kernel for conv int8 weight kernel sum

typer-J:
### Summary

One RVV kernel for CPU ConvInt8 resource initialisation, registered in the `CoreFunctions` table under `MNN_USE_RVV` with a scalar fallback when the pointer is null (non-RISC-V and RVV=OFF builds keep the original scalar path):

- `MNNConvInt8ComputeWeightKernelSum`: per-kernel int8 weight sum via `vle8` + `vwredsum` (i8->i16, 3 vector ops per chunk) instead of the previous widen-twice-then-`vredsum` chain (5 vector ops), walking each row in **e8m2** chunks so that on VLEN=128 one iteration covers 32 elements. A chunk sums to at most 32 * 128 = 4096, an eighth of the i16 range, so the one-step widening cannot overflow; scalar accumulation across chunks keeps the result exact. The SSE offset compensation moved from a per-iteration `#ifdef` block to a `compensateSseOffset` parameter.

This revision also drops `MNNUnpackConvScaleFromBuffer`, which the earlier version of this PR added. See "Why the unpack kernel was removed" below.

**The bias-float kernel that earlier revisions of this PR carried has been removed**, as requested in review. It is now proposed separately, as two alternative implementations (#4892 and #4893); see "Bias-float kernel removed" below for what changed and why the two alternatives exist.

Note for whoever merges: this PR and the two bias PRs each add one slot to the `CoreFunctions` table in `CommonOptFunction.h` and register it in `CommonOptFunction.cpp`, at the same place in that block. Git will report a conflict in those two files when a second one lands. It is an adjacency conflict -- both sides add a line, neither removes one -- so resolving it is a matter of keeping both additions.

### Bias-float kernel removed

Review asked for `MNNConvInt8ComputeBiasFloat` to come out of this PR, so that this PR covers the weight-sum kernel only. Removed in full: the kernel source, its `CoreFunctions` slot, its declaration and registration, its dispatch site, and the test section that covered it. The scalar loop in `CPUConvolution.cpp` is restored byte-for-byte as it was before this PR.

The reason the kernel measured at parity is the division, and it is worth stating because it is what the two follow-up PRs address. `outputScale` is loop-invariant, so the per-element `vfdiv` in `((bias * weightScale) * inputScale) / outputScale` can be replaced by a reciprocal computed once outside the loop. That was not done here because the pre-review revision precomputed the whole `inputScale / outputScale` ratio, which rounds differently and changes the result on every platform -- the concern that was raised in review, and correctly so. Taking the division out while keeping the original operation order is a separate change, and it is the change the two follow-up PRs make. The bias kernel's numbers are in those PRs rather than here.

### Changes in this revision

**1. The weight-sum kernel now walks e8m2 chunks instead of e8m1.**

The loop is charged per chunk, not per byte, and that is what the board data shows rather than an assumption. The `e8m1` column makes it visible: a 27-long row and a 32-long row cost the same (1.1708 us against 1.1707 us) even though the second carries 18.5% more bytes, because both are two chunks of 16. Over a longer interval the runtime ratio tracks the chunk count rather than the byte count -- from `kernelSize` 129 to 576 the chunk count goes 9 -> 36 (4.00x) and the byte count 4.47x, and the measured ratio is 3.88x, nearer the former. The remaining gain is therefore in the chunk count, not in the per-chunk instruction count -- a `vwadd`-style rewrite that lowers the instruction count per chunk but introduces an accumulator dependency chain measures *slower* than the form it would replace (53.25 us against 30.25 us at `kernelSize` 1152, and 3.72x slower than this revision's `e8m2`), so the structure is left alone and only the vector length moves.

e8m2 halves the chunk count on VLEN=128 (16 -> 32 elements per iteration): 1.25x on short rows, 2.11x at kernelSize 1152. The table is in "Board measurements" below.

**2. A claim in the previous description was wrong.** It said "A 32-lane i8 sum is bounded by 4096". The code at the time used `vsetvl_e8m1` / `vle8_v_i8m1`, i.e. 16 lanes on VLEN=128, so the real bound was 2048; disassembling the object confirmed the emitted `vsetvli` was `e8,m1`. The claim is true of this revision, and it is now stated with its derivation rather than asserted:

```
VLMAX(e8m2) * 128 <= 32767   =>   VLMAX <= 255   =>   VLEN <= 1020
```

At VLEN=128 a chunk sums to at most 4096, an eighth of the i16 range, and the bound only starts to bite above VLEN 1020. The accumulator across chunks is int32, so row length is unconstrained. e8m4 measures faster still (1.64x over e8m2 at kernelSize 1152) but its per-chunk cost grows, it lowers that bound to 510, and e8m8 gains nothing over e8m4 while occupying eight vector registers (1.00x at kernelSize 1152, 1.04x at 576, 0.99x at 129), so e8m2 is the stopping point. Both are in the table below.

**3. Integration-level verification.** Every number above comes from a kernel-only harness that does not link MNN: it proves the kernel, not that MNN calls it. I built the tree both ways (`MNN_USE_RVV=ON` and `=OFF`, `MNN_RVV_SPACEMIT_IME2=OFF`, `MNN_BUILD_TEST=ON`) and checked that the `CoreFunctions` slot really points at this kernel rather than merely being non-null -- the runtime slot address and the `nm` file offset of `MNNConvInt8ComputeWeightKernelSum_RVV` share the same low 12 bits, as a page-aligned load base requires. In the `MNN_USE_RVV=OFF` build the same slot is null and the call site falls through to the scalar loop.

(For anyone repeating this: `MNNGetCoreFunctions()` returns a null pointer before `MNNCoreFunctionInit()` has run, and that function is not exported, so a standalone probe cannot initialise the table -- it has to go through the test framework.)

**4. The `MNN_PUBLIC` commit is gone.** `MNN_PUBLIC` on `MNNGetCoreFunctions()` landed on master with #4847, so the `Export MNNGetCoreFunctions` commit (`b9d4b68f`) was dropped during the rebase.

### Board measurements (kernel only)

Board: riscv64, VLEN=128 (`rv64imafdcvh`), openEuler 24.03 SP3, GCC 14.3.1, 32 cores, `taskset -c 8` single-core pinning, load average 0.21 / 0.29 / 0.39 (1/5/15 min) recorded at the start of the run. Kernel-only harness, no MNN build; every arm compared for bit-exactness in the same run (**560 comparisons, 0 failures**).

All arms below are measured **in one run, in one process, from one build script**, so the table has a single baseline. Min over 15 rounds with the arm order rotated; the run-to-run spread is reported below, and a second full run reproduced every row to within 4%.

The `scalar` column is the loop compiled with `-march=rv64gc`: the object provably contains **zero** vector instructions (`objdump -d scalar.o | grep -cE '\bv[a-z0-9]+\.[a-z0-9]'` = 0), and that is the form the kernel replaces.

`kernelNum` = 256 throughout. Times are us.

| kernelSize | scalar | old | A | e8m1 | e8m2 | e8m4 | e8m8 |
|---|---|---|---|---|---|---|---|
| 1 | 0.8265 | 0.9427 | 0.9412 | 0.9406 | 0.9405 | 0.9408 | 0.9410 |
| 2 | 0.9400 | 0.9472 | 0.9480 | 0.9400 | 0.9399 | 0.9400 | 0.9401 |
| 3 | 1.5195 | 0.9469 | 0.9475 | 0.9399 | 0.9398 | 0.9400 | 0.9400 |
| 4 | 1.6344 | 0.9469 | 0.9473 | 0.9399 | 0.9398 | 0.9401 | 0.9401 |
| 9 | 2.1020 | 0.9466 | 0.9484 | 0.9378 | 0.9397 | 0.9396 | 0.9400 |
| 27 | 4.1955 | 1.7182 | 1.2489 | 1.1708 | 0.9380 | 0.9389 | 0.9383 |
| 32 | 4.7775 | 1.7900 | 1.2452 | 1.1707 | 0.9381 | 0.9393 | 0.9380 |
| 129 | 22.8757 | 6.8665 | 4.3042 | 3.6539 | 2.2117 | 2.0490 | 2.0798 |
| 576 | 77.0875 | 28.7520 | 23.6955 | 14.1620 | 7.3010 | 4.5545 | 4.3605 |
| 1152 | 151.4487 | 56.7195 | 53.2546 | 30.2510 | 14.3050 | 8.7175 | 8.7090 |

Ratios against the reference points that matter:

| kernelSize | e8m2 / scalar | e8m2 / old | e8m2 / e8m1 | e8m1 / old | e8m4 / e8m2 | e8m8 / e8m4 |
|---|---|---|---|---|---|---|
| 1 | 0.88x | 1.00x | 1.00x | 1.00x | 1.00x | 1.00x |
| 2 | 1.00x | 1.01x | 1.00x | 1.01x | 1.00x | 1.00x |
| 3 | 1.62x | 1.01x | 1.00x | 1.01x | 1.00x | 1.00x |
| 4 | 1.74x | 1.01x | 1.00x | 1.01x | 1.00x | 1.00x |
| 9 | 2.24x | 1.01x | 1.00x | 1.01x | 1.00x | 1.00x |
| 27 | 4.47x | 1.83x | 1.25x | 1.47x | 1.00x | 1.00x |
| 32 | 5.09x | 1.91x | 1.25x | 1.53x | 1.00x | 1.00x |
| 129 | 10.34x | 3.10x | 1.65x | 1.88x | 1.08x | 0.99x |
| 576 | 10.56x | 3.94x | 1.94x | 2.03x | 1.60x | 1.04x |
| 1152 | 10.59x | 3.97x | 2.11x | 1.88x | 1.64x | 1.00x |

Run-to-run spread (max-min over the 15 rounds, as a percentage of the min) is worth stating because it is not uniform across arms. At `kernelSize=1152`: `scalar` 13.35%, `old` 2.38%, `e8m1` 2.00%, `e8m2` 0.51%, `e8m4` 5.86%, `e8m8` 0.83%. The scalar loop is by far the noisiest arm, which is expected -- it is the slowest, so it is the most exposed to any interference on the core -- and it is the reason the min is used for the ratios rather than the mean.

Column meanings: `scalar` = the loop this kernel replaces (real scalar, 0 vector instructions). `old` = the unreleased RVV implementation this PR replaces (commit `1893b8d6d`, `vle8` + `vwcvt` + `vwcvt` + `vredsum` + `vmv.x`). `A` = the `vwadd`-style rewrite that lowers the per-chunk instruction count but introduces an accumulator dependency chain; it is slower than `e8m2` at every length and the gap widens with row length (1.01x at `kernelSize` 9, 1.33x at 27, 3.72x at 1152), which is why the structure was left alone and only the vector length moved. Its margin over `old` is thinnest at the longest rows (1.07x at 1152, against 1.60x at 129) -- the dependency chain costs it the advantage it starts with. `e8m1` = the form before this revision. `e8m2` = what this revision ships. `e8m4` / `e8m8` = the longer-VL variants kept in as the evidence for stopping at `e8m2`.

**Correction: the scalar column in the earlier version of this section was wrong.**

The previous revision of this description carried a `scalar` column reading **100.7651 us** at `kernelSize=1152`. That number was not the shipping scalar loop. It was the loop compiled with `-march=rv64gcv`, where GCC auto-vectorises it. To identify that rather than guess at it, the rerun built a control arm from the *same source* as the baseline with only the `-march` changed (`rv64gcv` instead of `rv64gc`); it measured **100.5786 us** on that row, agreeing with the old number to within 0.2%. The control is not in the table -- it is not a shipping baseline and keeping it would invite exactly the cross-column comparison this correction is about -- but it is what identifies the old number's origin.

The real scalar baseline at that row is **151.4487 us**, about 1.5x slower than the number the old table showed. The effect of the error is that the old table *understated* the speedup. Since a correction that makes the numbers look better deserves suspicion: the census that proves the baseline is scalar (`scalar.o` = 0 vector instructions, against 7 for the auto-vectorised control) is printed by the same script that produced the table, so the claim can be re-derived rather than taken on trust, and the script and its raw log are available on request.

The two tables that were here have been replaced by the single table above for one reason: they could not be compared with each other. Their `e8m1`/`new` columns were the same kernel measured in two different builds, and their scalar columns came from two different `-march` settings. One run, one baseline, no cross-table arithmetic.

**Short rows are a small regression, and it is real.**

At `kernelSize=1` the scalar loop is *faster* than every vector arm (0.8265 us against 0.9405 us, i.e. 0.88x). The cause is the fixed per-call cost of the vector path: the `vsetvlmax` and zero-register setup are paid once per kernel regardless of row length, and a one-element row has nothing to amortise them over. The crossover is sharp -- at `kernelSize=2` the two are level (0.9400 against 0.9399) and from `kernelSize=3` up the vector arm wins (1.62x).

This is worth stating because the earlier revision of this description claimed the kernel was "never slower than the code it replaces", which is true of `old` -- it is not true of the scalar loop. How far it matters in practice depends on the real `kernelSize`, which is `weightSize / outputCount` at the call site; a 1x1 convolution over a 1-channel input is the only shape that reaches `kernelSize=1`, and the absolute cost there is 0.11 us per layer. No scalar fallback for short rows was reintroduced: an interim revision of this PR had one, and it measured *slower* than the vector path at `kernelSize=9` (2.24x in favour of the vector arm), so a length threshold would have to sit at `kernelSize <= 1`, where the whole call is under a microsecond.

The gain grows with row length because that is where the chunk count grows; a row shorter than one vector has nothing to halve.

### Why the unpack kernel was removed

`MNNUnpackConvScaleFromBuffer` extracts the per-block quant scales out of the packed int8 weight buffer. Its unit of work is one contiguous copy of `UNIT * infoBytes` bytes, where `UNIT = GEMM_INT8_UNIT = 4` and `infoBytes = 4`, i.e. **16 bytes per block**. At VLEN=128 a single `vsetvl_e8m8` transfer already covers 128 bytes, so the loop always runs exactly one iteration on a partial vector: no implementation can beat `memcpy` here, and the measurement confirmed parity. Keeping it would have added a kernel plus a `CoreFunctions` slot for zero gain, so the scalar `unpackScaleFromBuffer` call is restored and the kernel, its table pointer and its dispatch are removed.

### Test

- Kernel-level unit test `test/backend/cpu/RVVConvInt8Test.cpp` (scalar oracle, exact bitwise match): **66 scalar comparisons pass** (`kernelNum` 1/4/9 x `kernelSize` 1/3/27/31/32/33/63/64/65/129/576 x `compensateSseOffset` off/on). `kernelSize` covers a single chunk (below one vector), several chunks, and a tail that is not a multiple of the vector length -- 31/32/33 straddle the first e8m2 chunk boundary and 63/64/65 the second.
- The stale claim in that file that `kernelSize` "spans the scalar fallback" was removed: that fallback no longer exists (dropped in the previous revision).
- Integration level, in a full MNN build: the kernel is installed in `CoreFunctions`, and an `MNN_USE_RVV=OFF` build still falls back to the scalar path with the slot null. Details under "Changes in this revision".

### Shared-library export

No longer carried here. `MNN_PUBLIC` on `MNNGetCoreFunctions()` was merged into master by #4847 (`CommonOptFunction.h:592`), so the commit that added it to this branch was dropped during the rebase. The tests still call `MNN::MNNGetCoreFunctions()` and link because master already exports it.


typer-J:
Summary of where this change stands.

## What it does

One kernel for the int8 convolution resource-initialisation path, registered through the function table inside the `supportRVV` branch with the scalar implementation kept as the default:

- `MNNConvInt8ComputeWeightKernelSum_RVV` — the kernel-sum loop inside `makeResourceInt8`.

## Performance (kernel microbenchmarks, native RVV VLEN=128, GCC 14.3.1)

| kernel | speedup |
|---|---|
| `MNNConvInt8ComputeWeightKernelSum` | 1.6x-1.9x on realistic row lengths; 1.05x at the smallest sizes, never slower |

The kernel-sum loop uses a single widening reduction (`vle8` + `vwredsum` i8->i16) instead of widening twice and then reducing. An interim revision of this PR added a scalar fallback for rows shorter than one vector; measurement showed that was a pessimisation (at `kernelSize=9` the scalar path is 1.2x slower than the vector path), so it was removed.

## Correctness

A scalar-oracle unit test (`test/backend/cpu/RVVConvInt8Test.cpp`) compares bitwise-exact across sizes that straddle one vector and leave non-multiple-of-VL tails; the kernel is bit-exact against the oracle on the board.

## Scope, stated honestly

This is a resource-initialisation helper that runs once per model load, not a per-inference `onExecute` hotspot, so the figure above is kernel-level and not an end-to-end claim.

Thank you for your time.

wangzhaode:
Thanks for narrowing the scope and clearly distinguishing initialization-time gains from inference performance! The 63 standalone kernel comparisons passed on our K3 board.

One small compatibility issue remains in CPUConvolution.cpp: the original bias expression is bias * weightScale * inputScale / outputScale, while the new scalar fallback uses bias * weightScale * (inputScale / outputScale). Precomputing the ratio changes floating-point rounding, including on non-RVV platforms. We confirmed differences with ordinary finite inputs; this is not evidence of a model-quality failure, but an RVV optimization should avoid changing the existing behavior on other platforms.

Could you please retain the original expression in the scalar fallback? Also, please use the original arithmetic as the test reference: the current reference already uses the precomputed ratio, so its bit-exact check does not establish equivalence to the old implementation. Passing inputScale and outputScale separately to the RVV helper is one straightforward option if you want to preserve the original operation order there too.

A small targeted comparison is enough; there is no need to expand the PR or add a large test suite. Thanks!

typer-J:
Rebased onto `a1526804` and addressed all three points.

**1. Op order restored.** The scalar fallback is byte-for-byte the original loop again (`bias * weightScale * inputScale / outputScale`), and `MNNConvInt8ComputeBiasFloat` now takes `inputScale` and `outputScale` separately instead of a precomputed ratio. I also removed the `scaleRatio == 1.0f` shortcut — you are right that the two roundings do not cancel when the scales are equal but not 1.0. For `(1.7f, 1.7f)` the scalar path and the `bias * weightScale` shortcut differ in 133 of 1000 samples.

**2. Oracle fixed.** `computeBiasFloatReference` now spells out the original expression instead of the ratio form, and the cases are `(inputScale, outputScale)` pairs. `(1.7f, 1.7f)` is included specifically for the shortcut case, plus non-representable pairs and a zero on either side.

**3. `MNN_PUBLIC` commit dropped.** It landed on master with #4847, so `b9d4b68f` was dropped during the rebase.

`mergeable` is now `true`. Evidence is in the PR description, including a native check showing the kernel matches the fixed oracle on all 35 cases while the pre-review form differs on 1241 samples.

One thing to flag: the bias-float timings in the description were measured with the shortcut in place, so they no longer describe the current kernel, and I do not have board access in this revision. The reorder adds one `vfdiv` per element; the scalar column is the closer reference and the kernel was never slower than it.

wangzhaode:
Thanks for restoring the original operation order, fixing the oracle, and clearly flagging that the old bias-kernel timings no longer apply!

The remaining follow-up is to check performance of the current bias kernel after removing the shortcut. When board access is available, could you please provide a small scalar-versus-RVV A/B for a few representative channel counts and update the timing note accordingly? Please use the current arithmetic on both sides and include the basic board/build context. There is no need for an end-to-end model benchmark or a large sweep.

Once that check is complete and you consider the patch ready, please mark the PR as ready for review so we can finish the review. Thanks again!

typer-J:
Board access is back, so here is the targeted comparison you asked for — and it changes one of the numbers in the description.

**Setup.** riscv64, VLEN=128, openEuler 24.03 SP3, GCC 14.3.1, single core pinned with `taskset`. Kernel-only harness, no MNN build; all arms bit-exact against each other within the same run. Three arms, because the code the kernel replaces is compiled without the V extension:

- `NoV` — the scalar loop built for the baseline ISA (`-march=rv64gc`; `objdump -d` confirms zero vector instructions in the object). This is the arm the kernel actually replaces.
- `Auto` — the same loop at `-march=rv64gcv`, so the auto-vectoriser is free to act (15 vector instructions).
- `RVV` — the hand-written kernel from this PR.

**Result.** With the original operation order restored the division is back inside the loop:

| mode | size | NoV | Auto | RVV | RVV/NoV |
|---|---:|---:|---:|---:|---:|
| asymmetric, no scaling | 64 | 0.05 us | 0.02 us | 0.01 us | **5.04x** |
| | 4096 | 2.87 | 1.07 | 0.71 | **4.04x** |
| symmetric, scaled | 64 | 0.18 | 0.18 | 0.18 | 0.96x |
| | 4096 | 11.18 | 11.65 | 11.64 | 0.96x |

The scaled path lands at parity, about 4% behind, and it is the division that sets the pace: roughly 8.3 us of the 11.2 us at size 4096 is the divide, while the unscaled path is 2.9 us for the same loop with the division removed. Vectorising the divide does not pay on this core — the auto-vectorised arm lands at 11.65 us, the same as the hand-written one — so this is not an artefact of how the intrinsic is written.

**This retires the earlier "never slower than it" claim**, which was measured on the precomputed-ratio form with no division in the loop. I have updated the description to say so plainly rather than leave the old number standing: RVV is **4.0-5.0x** faster when the bias path is unscaled, and **at parity (~4% behind)** when it is scaled.

Since the scaled case is the common one for symmetric quantisation, a `CoreFunctions` slot whose common path sits at parity may not be worth carrying. It is your call — if you would rather not, the bias kernel, its table slot and its dispatch can come out of this PR the same way `MNNUnpackConvScaleFromBuffer` did. Tell me which way you want it and I will adjust before the final review.

`MNNConvInt8ComputeWeightKernelSum` is unaffected by this and stays at 1.6-1.9x on realistic row lengths, never slower than the code it replaces.

Marking this ready for review — all three points from your comment are in, and the open question is only whether the bias kernel is worth carrying in the scaled case.

typer-J:
A short summary now that the board numbers are in, so all three of your points are closed off in one place.

**1. The original expression is retained.** The scalar fallback is byte-for-byte the original loop again — `bias[i] * weightScale[i] * mInputScale / mOutputScale`, with no precomputed ratio.

**2. The oracle uses the original arithmetic.** `computeBiasFloatReference` spells the expression out, and the cases are `(inputScale, outputScale)` pairs including `(1.7f, 1.7f)` — the pair that catches an implementation which precomputes the ratio and skips the scaling step.

**3. `inputScale` and `outputScale` are passed separately** to the RVV helper, as you suggested, and the `inputScale == outputScale` short-circuit is gone.

**Board re-measurement.** With the operation order restored the division is back inside the loop, and that changes the conclusion for `MNNConvInt8ComputeBiasFloat`:

| mode | size 4096 | NoV | Auto | RVV | RVV/NoV |
|---|---:|---:|---:|---:|---:|
| asymmetric, no scaling | 4096 | 2.87 us | 1.07 | 0.71 | **4.04x** |
| symmetric, scaled | 4096 | 11.18 | 11.65 | 11.64 | **0.96x** |

The scaled path — the common one for symmetric quantisation — lands at parity, and the division is what sets the pace (~8.3 of the 11.2 us). The auto-vectorised arm lands at the same 11.65 us, so this is not an artefact of how the intrinsic is written. I have retired the earlier "never slower" claim in the description rather than leave the old numbers standing.

`MNNConvInt8ComputeWeightKernelSum` is unaffected and stays at 1.6-1.9x on realistic row lengths.

Since parity on the common path may not be worth a function-table slot, I would rather ask than guess: would you prefer the bias kernel, its slot and its dispatch removed from this PR the same way `MNNUnpackConvScaleFromBuffer` was? Either way is fine — tell me which and I will adjust.

Looking forward to your latest thoughts.

wangzhaode:
Thanks for restoring the original arithmetic and providing the updated board measurements! Yes, please remove `MNNConvInt8ComputeBiasFloat_RVV`, its function-table slot, declaration and registration, and the corresponding dispatch from this PR, leaving the original scalar bias loop in place. Please also remove the tests that only exercise that helper.

The unscaled case does benefit, but the common scaled path is essentially at parity, so we would prefer to keep this resource-initialization change focused on `MNNConvInt8ComputeWeightKernelSum`, which has a clearer measured benefit. Please retain that kernel and its correctness coverage.

A focused build and the retained weight-sum tests are sufficient after this cleanup; no new model benchmark or broad performance sweep is needed. Thanks again for the careful follow-up!

typer-J:
**A note on why the scaled path was at parity, before the removal lands**

You asked me to remove the bias kernel because the common scaled path measured at parity. Before I do, I checked whether that parity is inherent to the loop or caused by the instruction it contains. It is the latter, and there are two ways to remove it. Since this changes the decision you already made, I would rather show you the numbers than act on them.

**Why the loop is slow.** `outputScale` is loop-invariant. The current kernel evaluates `((bias * weightScale) * inputScale) / outputScale`, so every element pays a `vfdiv`. The division is ~74% of the loop at size 4096 (11.2 µs total, 2.9 µs with the division removed). Because the divisor never changes, the division can be replaced by a reciprocal computed once, plus a correction:

```
r  = 1.0f / outputScale        // hoisted, once
q0 = t * r                     // per element
e  = fma(-outputScale, q0, t)  // exact residual
q  = fma(r, e, q0)             // corrected quotient
```

**Board measurements** (riscv64, VLEN=128, GCC 14.3.1, `taskset -c 8`, standalone kernel harness, no MNN build; every arm bit-compared in the same run):

| arm | size 4096 | vs NoV scalar |
|---|---:|---:|
| NoV — scalar loop built for `rv64gc`, the code the kernel replaces | 11.20 µs | 1.00x |
| current kernel (one `vfdiv` per element) | 11.67 µs | 0.96x |
| reciprocal + Markstein correction | **1.78 µs** | **6.29x** |
| reciprocal only | **1.19 µs** | **9.33x** |

`objdump -d` confirms the NoV object contains zero vector instructions, so the baseline is genuinely scalar.

**The two differ in accuracy.** Against the original expression as reference, over 200 scale pairs x 819200 elements:

| arm | elements differing in the last bit | max relative error |
|---|---:|---:|
| reciprocal only | 24.4% | 6.2e-06 |
| reciprocal + Markstein | **0.000%** | 0 |

On realistic quantisation scale pairs (`is=1.7 os=3.0`, `is=0.375 os=0.375`) the reciprocal-only arm differs on 32-33% of elements by 1 ULP, while the corrected arm is bit-identical on every finite value.

**One safety point that needs a guard.** `1.0f / outputScale` overflows to `inf` when `outputScale` is subnormal, which the original per-element division does not do. A check hoisted out of the loop fixes it:

```cpp
const bool rcpOk = std::isfinite(r) && r != 0.0f && fabsf(r) >= 1.17549435e-38f;
// otherwise fall back to vfdiv_vf
```

With the guard, the subnormal and overflow cases produce no non-finite values at all, and the timing is unchanged (6.29x vs 6.33x).

**Caveats, stated plainly.** This is a standalone kernel harness, not an MNN build, so the numbers are kernel-level. I verified bit-exactness on finite values only; the subnormal region is covered by the guard rather than by an equivalence proof. I have not wired either arm into the tree.

**So, three options — your call:**

1. **Neither.** Remove the kernel as you asked, and this becomes a note for a possible future PR. This is a perfectly good outcome and I am fine with it.
2. **Reciprocal + Markstein.** 6.29x with no change to any finite output, so no cross-platform behaviour change.
3. **Reciprocal only.** 9.33x, but it changes the last bit on ~25-33% of elements, which is the same class of change you flagged on the earlier precomputed-ratio revision. I would not propose this one on its own.

My own preference is 1 or 2 — not 3, for the reason above. But if you would rather keep the resource-initialisation change minimal and not carry this at all, option 1 is entirely reasonable and I will proceed with the removal.

PR #4425, opened 2026-05-06
Feature/rvv support cpu

Sherlockzhangjinge:
## Description

Optimize functions below with RVV:
MNNBinaryMaxInt8
MNNBinaryMinInt8
MNNBinaryMulInt8
MNNBinarySqdInt8
MNNBinarySubInt8
_ArmBasicMNNPackC4ForMatMul_A
_MNNPackC4Int8ForMatMul_ASparse
MNNScaleAndAddBiasInt8
MNNSumByAxisLForMatmul_A

## Module

CPU
## Type

- [x] Feature
- [ ] Bugfix
- [ ] Perf
- [ ] Refact
- [ ] Style
- [ ] Doc
- [ ] Test
- [ ] Chore

## Checklist

- [x] Commit message follows `[Module:Type] Description` format
- [x] Code compiles without errors
- [x] Tested on relevant platform(s)
- [x] No unrelated format or style changes included

PR #4867, opened 2026-09-12
[CPU:Perf] Optimize int8 Min/Max/ScaleAndAddBias with RVV

typer-J:
## Description

Focused follow-up to the closed PR #4425, rebased on latest master, implementing the reviewer's suggestion to split it into smaller, focused PRs with correctness tests and performance A/B results.

This PR adds RVV 1.0 intrinsics implementations for two int8 elementwise kernels with stable, measured gains on a riscv64 RVV board:

- `MNNBinaryMinInt8` (+2.31x kernel-level)
- `MNNBinaryMaxInt8` (+2.26x kernel-level)

### Update after review: ScaleAndAddBias removed

`MNNScaleAndAddBiasInt8` was part of the original revision and has been **removed** from this PR. Thank you for the pack=4 measurement — it is correct, and the reason is visible in the kernel structure.

The CPU path calls `MNNScaleAndAddBiasInt8` with `pack = core->pack`, which is 4 on the int8 path (`CPUScaleInt8.cpp`, `biasNumber = 1`). The RVV kernel vectorised **along `pack`**: at `pack = 4` the inner loop runs with `vl = 4` while the 32-bit vector registers it feeds are 16 lanes wide at VLEN=128, and `alpha`/`bias` — which are per-channel and therefore loop-invariant across planes — were reloaded once per plane. The result is exactly what you measured.

Removed in this revision:
- `source/backend/cpu/riscv/rvv/MNNScaleAndAddBiasInt8.cpp` (whole file)
- its `extern` declaration in `Int8FunctionsOpt.cpp`
- the runtime dispatch block in `MNNScaleAndAddBiasInt8()`

The scalar body of `MNNScaleAndAddBiasInt8` is byte-identical to master, so this revision is a no-op for that kernel.

Scale can follow separately once it benefits pack=4 — the natural direction is what you suggested, vectorising across spatial positions instead of across channels, so that `alpha`/`bias` become loop-invariant scalars.

Kernels from #4425 without a stable gain remain excluded: the four float-path binary ops (Add/Sub/Mul/Sqd, ~1.0-1.2x, noise range) and `_MNNPackC4Int8ForMatMul_ASparse` (0.89x, regression). `MNNSumByAxisLForMatmul_A` was already correctly registered on master and is untouched.

### Key implementation points

- Integer Min/Max use a fixed-point path; division by 2^16 is done with an add + select + `vsra` sequence that is bit-exact with C++ truncating integer division (the scalar reference's `roundf` on Min operates on an already-truncated integer, so it is a no-op). This addresses the reviewer's point about avoiding `vdiv`, which is emulated and very slow on RVV hardware.
- No VLA usage anywhere (addresses #4425 review point 3); all loops are VLEN-independent via `vsetvl` tail handling.
- Neither kernel has a `CoreInt8Functions` slot, so there is no table entry that could be rewritten at run time. Dispatch is therefore an inline `supportRVV` test at the top of each scalar body in `Int8FunctionsOpt.cpp`, inside `#ifdef MNN_USE_RVV`; the scalar body stays compiled unconditionally, so an `MNN_USE_RVV=ON` binary still runs correctly on a CPU without the V extension. The guards are narrow and do not touch sibling functions in the same block (the overly broad guards in #4425 caused 53 build errors; none occur here).

## Response to PR #4425 review comments

| # | Review comment (wangzhaode) | Status in this PR |
|---|---|---|
| 1 | Missing semicolon in registration | No registration table touched; dispatch is an inline runtime test, no missing-statement patterns |
| 2 | `#elif MNN_USE_RVV` should be `defined()` | Only `#ifdef` / `#ifndef` / `#ifdef MNN_USE_RVV` used |
| 3 | VLA in sparse packing | No VLA; sparse packing excluded from this PR (measured regression, 0.89x) |
| 4 | Use shift instead of `vdiv` | `truncDivPow2` add/select/`vsra` sequence, bit-exact with C++ division |
| 5 | Missing performance benchmarks | Provided below (kernel-level A/B, 20 fresh processes) |
| 6 | Checklist unchecked | Checked below after board verification |
| 7 | Missing trailing newline | All new files end with a newline |

## Verification

Board: riscv64 (`rv64imafdcvh`), openEuler, g++ 14.3.1, Ninja, 32 cores, `taskset -c 2` single-core pinning.

**Build matrix** (both full `libMNN.a` builds, exit 0):
- `-DMNN_USE_RVV=ON` (Ninja, Release): links OK
- `-DMNN_USE_RVV=OFF` (Ninja, Release): links OK — non-RVV builds unaffected

**Correctness** (standalone kernel test, RVV vs `no-tree-vectorize` scalar oracle, 20 fresh processes, all PASS):
- Min/Max: 7 sizes (1..4099, incl. tails) x 3 broadcast modes, non-trivial zero points, signed inputs, clamp bounds

**Performance** (kernel-level microbenchmark, single core, 20-process median; scalar = `no-tree-vectorize` reference compiled with the same `-O3 -march=rv64gcv`):

| kernel | scalar (us) | RVV (us) | speedup | range |
|---|---|---|---|---|
| MNNBinaryMinInt8 (4099 elems, broadcast) | 12.75 | 5.51 | **2.31x** | 2.07-2.44x |
| MNNBinaryMaxInt8 (4099 elems, broadcast) | 12.54 | 5.49 | **2.26x** | 2.05-2.31x |

These are kernel-level numbers, not end-to-end.

## Module

CPU

## Type

- [ ] Feature
- [x] Perf
- [ ] Bugfix
- [ ] Refact
- [ ] Style
- [ ] Doc
- [ ] Test
- [ ] Chore

## Checklist

- [x] Code compiles without errors (verified with `MNN_USE_RVV=ON` and `=OFF` on riscv64)
- [x] Tested on relevant platform(s) (riscv64 RVV board; correctness + 20-process A/B benchmark)


wangzhaode:
Thanks for splitting this work into a focused PR and providing the measurements! We checked it on K3 (VLEN=256, GCC 15.2): 600 Min/Max comparisons and 112 Scale comparisons matched the original scalar implementations, and Min/Max also showed clear speedups.

There is one performance issue with ScaleAndAddBias at the pack=4 layout used by the CPU path. In a single-core, same-process kernel A/B test (-O3, scalar reference compiled for rv64gc and the RVV implementation for rv64gcv), pack=4, planeNumber=257, biasNumber=1, shift=16 took about 4.64 us with the original implementation versus 10.32 us with this PR. Repeated measurements were consistent. These are kernel timings, not end-to-end model results.

The current inner loop vectorizes only four channels at a time, so the gains measured with larger pack values do not carry over to this layout. To keep this PR straightforward, could you please drop the ScaleAndAddBias change for now and retain the Min/Max improvements? Scale can follow separately once it benefits pack=4, for example by processing multiple spatial positions together. A small targeted A/B check is sufficient; no extensive benchmark suite is needed. Thanks again!

typer-J:
Rebased onto `a1526804`; the ScaleAndAddBias kernel is gone.

Removed: `rvv/MNNScaleAndAddBiasInt8.cpp`, its `extern` declaration, the dispatch block at the top of `MNNScaleAndAddBiasInt8`, the test accessor, and the description in the PR body. `MNNScaleAndAddBiasInt8` is back to the scalar body it had before this PR, and Min/Max are untouched.

The conflict during the rebase was in `Int8FunctionsOpt.cpp`, in the `#ifdef MNN_USE_RVV` extern block that #4838 also edited; the resolution keeps both sides.

`mergeable` is now `true` — 3 files, +169/-0.

wangzhaode (approved):

PR #4869, opened 2026-09-13
[CPU:Bugfix] Preserve RVV INT8 sum accumulator tails

luke-opus:
## Description

Fix incorrect INT8 input sums in the general `MNNSumByAxisLForMatmul_A_RVV` path when tail-agnostic instructions overwrite inactive accumulator lanes. For example, with VLEN=128, 17 input channels, two output positions and scale=1, the original kernel returns `[-13, -11]` instead of `[17, 34]` under an ISA-permitted tail policy.

The accumulators are initialized at VLMAX and reduced at VLMAX, but the two additions use the current, potentially smaller VL. Change both additions to `__riscv_vadd_vv_i32m4_tu` so inactive lanes retain zero or contributions from earlier channel blocks. The final full-width reduction remains necessary to include those earlier contributions.

Add `RVVSumByAxisTailTest` with eight packed-input cases covering aligned and partial channel blocks, paired and unpaired output positions, and the separate single-position fast path. The test checks an independent exact reference, poisoned input padding and output guards. It supports both the registered `backend/cpu/rvv/sum_by_axis_tail` test and a standalone entry point. Reproduction commands are in `test/backend/cpu/RVVSumByAxisTailTest.md`.

## Module

CPU / standard RISC-V RVV

## Type

- [ ] Feature
- [x] Bugfix
- [ ] Perf
- [ ] Refact
- [ ] Style
- [ ] Doc
- [ ] Test
- [ ] Chore

## Validation

Compiled and executed the final standalone test against both the original kernel at `a03b005cf6f888ebf092e4753840f935827f9c36` and the corrected kernel. Environment: Zig 0.15.2 / Clang 20.1.2, RV64IMAFDCV, `-O2`, static RISC-V Linux binaries, Windows QEMU 11.1.0 TCG system emulation, Debian Linux 6.12.107. Automatic vectorization and LTO were disabled; the kernel's RVV intrinsics were compiled normally.

| VLEN | `rvv_ta_all_1s` | Original kernel | Corrected kernel |
|---|---|---|---|
| 128 | true | 5/8 cases fail | 8/8 pass |
| 256 | true | 7/8 cases fail | 8/8 pass |
| 128 | false | 8/8 pass | 8/8 pass |
| 256 | false | 8/8 pass | 8/8 pass |

`rvv_ta_all_1s=true` is essential for exposing this bug: it selects an ISA-permitted all-ones tail result. With it disabled, QEMU preserves the tail and the original kernel passes these cases. At VLEN=256, C=16/E=2 is another observed failure: `[0, 16]` instead of `[16, 32]`.

Also compiled the test's suite-registration branch as a RISC-V object, checked that the non-RVV build excludes the implementation, and ran `git diff --cached --check`. The full MNN test executable, operator/model regressions and physical RISC-V hardware have not been run; these results establish correctness for the tested kernel paths, not hardware performance.

## Checklist


wangzhaode:
Thanks for the fix and the clear tail-policy analysis! The production change looks correct, and all eight standalone cases passed on our K3 board (VLEN=256).

Could you please add a small runtime guard to the registered test before reading the vlenb CSR? MNN_USE_RVV only describes the build configuration; when supportRVV is false, the test should skip without executing the CSR read or any RVV instructions. Otherwise, an RVV-enabled test binary can still raise an illegal-instruction exception on a CPU without the V extension.

This should be a small test-only adjustment; no broader changes are needed. Thanks!

PR #4874, opened 2026-09-14
[CPU:Perf] Optimize C4 bilinear sampler with RVV

typer-J:
## Description

Adds an RVV 1.0 intrinsics implementation of the C4 bilinear sampler, `MNNSamplerC4Bilinear`, measured at **2.55x kernel-level** on a riscv64 RVV board at the real call-site configuration.

The scalar body costs **38.9 ns per output pixel**: 16 byte loads (four neighbours x four channels), plus six rounding operations — two `ceilf` to pick the neighbour coordinates and four `roundf` for the output. GCC expands each of those inline into a short `fcvt`/`fabs`/`fsgnj` sequence rather than calling into libm (the generated object file contains no `@plt` call sites at all), so this cost is instruction count, not call overhead; one term of the blend even carries a `double` intermediate. That is enough scalar work per pixel to pay for the vector setup, and the vector body brings it down to 15.2 ns per pixel.

### Implementation notes

- **`vl` is fixed at 4**, the C4 channel count, so the whole call runs on a single `vsetvl` with no per-pixel reset. At VLEN=128 `e32m2` holds 8 lanes, so the four channels leave the register group half empty by design.
- **LMUL stays at `e8mf2` / `e32m2`**, deliberately not the widest available grouping. This is the same choice the existing `MNNAbsMaxFP32` (m1), `MNNFloat2Int8` (m2) and `MNNSumByAxisLForMatmul_A` (m4) kernels make.
- **`roundAndPack` truncates and compares the exact fraction** (`vfcvt_rtz` plus `vmfge` against `0.5f`) instead of adding `0.5f` before truncating. The addition rounds by itself and can push a value sitting just below a tie onto the far side, which would diverge from `roundf()` exactly on a rounding boundary.
- **The output is not bit-identical to the generic body, and this revision no longer claims it is.** The kernel evaluates the blend in FP32 throughout, while `_sampleBilinearCommon` computes the c10 term as `yF * (1.0 - xF) * c10` with `1.0 - xF` in double. The two agree except on inputs that land exactly on a rounding boundary, where the FP32 result is one grey level lower:

  | | value | `roundf` |
  |---|---|---|
  | generic (`1.0 - xF` in double), `xF = 0.3701782822608948`, `yF = 0.26382631063461304`, `(c00,c01,c10,c11) = (193,161,211,52)` | 173.5 | 174 |
  | kernel (FP32) | 173.4999847 | 173 |

  The four weights keep the scalar expression order, which is what keeps the divergence at that single step. The kernel is deliberately **not** switched to double just to preserve the boundary; the case above is pinned in the test instead.
- A `vl < 4` guard falls back to the scalar body: a core too narrow for four `e8mf2` lanes would otherwise drop channels from every pixel.

### Dispatch

`MNNSamplerC4Bilinear` is reached through the function table. `ImageProcessUtils::choose(ImageFormat, Filter, bool)` returns `coreFunctions->MNNSamplerC4Bilinear` for the RGBA/BGRA bilinear case, and the pipeline calls through that pointer. Registering the slot in the existing `if (gCoreFunction->supportRVV)` branch of `MNNCoreFunctionInit` is therefore the *entire* wiring, and the kernel is a drop-in replacement exactly like the neighbouring `MNNDeconvRunForUnitDepthWise` registration.

The scalar body of `MNNSamplerC4Bilinear` is untouched, and every change is inside `#ifdef MNN_USE_RVV`, so an `MNN_USE_RVV=ON` binary still runs correctly on a CPU without the V extension.

### Incidental fix

`test/CMakeLists.txt` had four `set_property(` calls sharing a single closing parenthesis, which makes `MNN_BUILD_TEST=ON` fail to configure. It is fixed here in the same multi-argument form the four calls obviously intended, and the new test file is added to that list. Without it the test below cannot be built at all; I am happy to split it out if you would prefer it as a separate commit.

## Verification

Board: riscv64 (VLEN=128, `zve64d`), openEuler 24.03 SP3, g++ 14.3.1, Ninja, 32 cores, `taskset -c 8` single-core pinning.

**Build matrix** (`MNN_BUILD_TEST=ON`, Ninja, Release):
- `-DMNN_USE_RVV=ON`: full `libMNN.a` + `run_test.out`, 0 errors / 0 warnings
- `-DMNN_USE_RVV=OFF`: configures and builds cleanly — non-RVV builds unaffected

**Correctness**: `run_test.out backend/cpu/rvv/image_process` — **52 cases passed** (`supportRVV=1`). The test drives the function table, not the `_RVV` symbol, so it exercises the registration rather than merely the kernel: it asserts `core->MNNSamplerC4Bilinear == MNNSamplerC4Bilinear_RVV` and then calls through that slot. 17 lengths (0, 1, 2, 3, 4, 7, 15, 16, 17, 31, 32, 33, 63, 64, 65, 127, 129) x 3 point sets, plus the rounding-boundary case above.

The comparison against the oracle uses an **explicit one-level tolerance on pixel values**, because the oracle reproduces the generic body's double intermediate while the kernel is FP32. The margin after the written range stays **byte-exact**, so a write past the requested range still fails rather than being absorbed by the tolerance. The boundary case is what makes the tolerance meaningful: it is the only case in the suite that exercises the one-level difference, and a kernel that drifted further, or reordered the weights, would still fail.

I checked the tolerance on the host before pushing by evaluating both arithmetic sequences with FP contraction off: across all 52 cases the boundary case is the only one with a difference, four bytes (one per channel), and no byte differs by more than one.

Regression: `backend/cpu/rvv` 3/3 pass, `cv` 20/20 pass.

**Performance** (kernel-level microbenchmark, single core, fresh process, 5 rounds with arm order alternating, min over rounds; generic arm is the untouched scalar kernel running in the same process, same binary, same input):

| shape | generic (ns/call) | RVV (ns/call) | speedup | jitter |
|---|---|---|---|---|
| identity dx=1, ow=512 | 19918 | 7803 | **2.553x** | 0.2% |
| identity dx=1, ow=1920 | 74659 | 29419 | **2.538x** | 0.0% |
| scale 2x dx=0.5, ow=512 | 19919 | 7803 | **2.553x** | 0.3% |
| scale 2x dx=0.5, ow=1920 | 74748 | 29379 | **2.544x** | 0.2% |
| skew dy=0.5, ow=512 | 20209 | 7814 | **2.586x** | 0.3% |
| skew dy=0.5, ow=1920 | 88080 | 36948 | **2.384x** | 4.9% |

`ow=512` is a single live `CACHE_SIZE` tile; `ow=1920` is 4 tiles (3x512 + 384) and is included only to show the tiling loop does not change the picture.

These are kernel-level numbers, not end-to-end. In the pipeline this kernel sits in `transformImage`, which is where the 38.9 -> 15.2 ns per pixel applies.

**Output equivalence.** The timing harness only compares a strided sum of the destination buffer, which is a smoke test rather than an equivalence proof, so the two arms are additionally compared directly: same process, same input, both arms driven through the function-table slot, and the sentinel-filled margin after the destination range checked for exact equality, so a write past the requested range would be caught. **12804 cases / 6124464 bytes across those geometries produced no difference.** Coverage: geometries from 3x2 to 1920x1080, strides with and without row padding, output widths 1..1920, `count` up to a full `CACHE_SIZE`=512 tile, four sample shapes (identity, 2x scale, skewed, off-grid fractional), and a border battery that drives coordinates onto and past every edge with negative and non-binary-representable steps.

Those inputs happen not to land on a rounding boundary, so they agree byte for byte. The inputs that do land on one are pinned as the registered test case described above; that case is where the one-level divergence is reproduced deliberately.

## Module

CPU

## Type

- [ ] Feature
- [x] Perf
- [ ] Bugfix
- [ ] Refact
- [ ] Style
- [ ] Doc
- [ ] Test
- [ ] Chore


wangzhaode:
Thanks for the RVV optimization and the detailed validation! The overall implementation looks reasonable.

One small point: the output is not strictly bit-identical to the generic implementation. The generic expression contains a double intermediate in `yF * (1.0 - xF) * c10`, while the new kernel uses FP32 throughout. A local C++ reproduction of the two arithmetic sequences (with FP contraction disabled, not a board-level RVV test) gives this rounding-boundary example:

- `xF = 0.3701782822608948f`, `yF = 0.26382631063461304f`
- `(c00, c01, c10, c11) = (193, 161, 211, 52)`
- Generic: `173.5 -> 174`; new FP32 arithmetic: `173.4999847 -> 173`

There is no need to switch the entire kernel to double just to preserve this boundary behavior. Could you please update the bit-identical claims in the description/code comments and add a small rounding-boundary case to the existing test, with an explicit one-level tolerance for pixel values? Please keep the sentinel checks exact. No larger test suite or additional benchmark report is needed.

Thanks again!

typer-J:
Thanks for the precise pointer — the rounding-boundary example made the fix unambiguous.

**1. The claims are updated.** The description and the kernel comment now state plainly that the output is *not* bit-identical to the generic body, with the reason (`1.0 - xF` in double there, FP32 throughout here) and the one-grey-level example from your message.

**2. A rounding-boundary case was added** to the existing test, pinned with exactly the inputs you gave: `xF = 0.3701782822608948`, `yF = 0.26382631063461304`, `(c00,c01,c10,c11) = (193,161,211,52)`. It is wired into the same test entry, so the tolerance is exercised rather than merely declared.

**3. The tolerance is one level, and only on pixels.** `|diff| <= 1` applies to the written pixels only. The margin after them is still compared byte-exact against its `0x5a` sentinel, so a write past the requested range fails instead of being absorbed by the tolerance. A kernel that drifted further, or reordered the weights, still fails.

**4. Real board run.** Built on the riscv64 board with `MNN_BUILD_TEST=ON -DMNN_USE_RVV=ON` (static, Ninja, Release, `-DMNN_BUILD_SHARED_LIBS=OFF`) and ran the registered case:

```
RVV image-process: 52 cases passed (supportRVV=1)
TEST_CASE_AMOUNT_UNIT: {"blocked":0,"failed":0,"passed":1,"skipped":0}
```

52 = 17 output lengths x 3 point sets + the boundary case. The test drives the function-table slot and asserts `core->MNNSamplerC4Bilinear == MNNSamplerC4Bilinear_RVV`, so it exercises the registration rather than only the kernel. `-DMNN_USE_RVV=OFF` also configures and builds cleanly.

The host-side check you asked for is in as well: across all 52 cases the boundary case is the only one that differs — four bytes, one per channel, none by more than one.

Looking forward to your latest thoughts.

wangzhaode (approved):

PR #4881, opened 2026-09-16
[App:Bugfix] Fix chat list flicker during streaming

### LINKED ISSUE #4813: 模型在流式输出时文本会上下跳动
【bug-哔哩哔哩】 https://b23.tv/XapZ4dv

### LINKED ISSUE #4862: 模型对话创建表格bug
模型画表格时，你们对话页面会闪屏，频繁上下运动（闪烁）

ValiantDiligent:
## Description

Fix chat-list flicker / jump during LLM streaming in the Android MnnLlmChat app.

Fixes  #4862  , Fixes #4813
Root causes addressed:
- **Scroll jumps**: `LinearLayoutManager` re-anchored to position 0 whenever a streaming
  item's height changed. Set `stackFromEnd = true` so it anchors to the last item, and
  replace the unreliable `computeVerticalScrollRange()` estimate for "distance from bottom"
  with the actual last-item view position.
- **Double view binding**: the payload `onBindViewHolder` called `super`, which forwarded to
  the full-bind overload and re-ran a complete rebind on top of the incremental one. Now it
  dispatches explicitly, and binding is split into `applyImageAndLoading()` /
  `applyMetricsAndActions()` so the streaming path renders metrics & actions correctly.
- **Markdown table reflow**: Markwon draws a whole table as `TableRowSpan`s in one `TextView`,
  needing two layout passes per appended row, so tables visibly reflowed while streaming. Added
  `MarkdownTableView`, which renders one native View per row and appends incrementally. Hardened
  GFM parsing (`parseTableRows` / `alignSeparator`) so malformed delimiter rows (models often emit
  one cell too few) no longer degrade the table into plain text.
- **Theme color**: `UiUtils.getThemeColor()` returned a resource id instead of an ARGB value for
  reference attributes; fixed, and added a shared `dpToPx()` helper.

Adds JVM unit tests for table parsing in `MarkdownBlockParserTest`.

## Module

App (Android · MnnLlmChat)

## Type

- [ ] Feature
- [x] Bugfix
- [ ] Perf
- [ ] Refact
- [ ] Style
- [ ] Doc
- [x] Test
- [ ] Chore

## Checklist


wangzhaode:
Thanks for fixing the streaming flicker and for completing the CLA! The changes to avoid duplicate binding look useful.

There is one small table-parsing regression to address before merging: `splitCells()` currently splits on every `|`, including an escaped pipe (`\|`) inside a cell. For example:

```markdown
| Content | Note |
| --- | --- |
| a\|b | normal |
```

This should produce two cells, `a|b` and `normal`. The current implementation splits the data row into three cells, then `parseTableRows()` truncates it to the header's two columns, dropping `normal`.

Could you please split only on unescaped separator pipes, preserving escaped pipes as cell content, or reuse the existing Markdown parser's table handling? One focused regression test for this example is sufficient; no broad new test suite is needed. Table alignment styling can be handled separately and does not need to block this fix.

Thanks again!

ValiantDiligent:
> Thanks for fixing the streaming flicker and for completing the CLA! The changes to avoid duplicate binding look useful.
> 
> There is one small table-parsing regression to address before merging: `splitCells()` currently splits on every `|`, including an escaped pipe (`\|`) inside a cell. For example:
> 
> ```
> | Content | Note |
> | --- | --- |
> | a\|b | normal |
> ```
> 
> This should produce two cells, `a|b` and `normal`. The current implementation splits the data row into three cells, then `parseTableRows()` truncates it to the header's two columns, dropping `normal`.
> 
> Could you please split only on unescaped separator pipes, preserving escaped pipes as cell content, or reuse the existing Markdown parser's table handling? One focused regression test for this example is sufficient; no broad new test suite is needed. Table alignment styling can be handled separately and does not need to block this fix.
> 
> Thanks again!

Thanks — confirmed and fixed.

Reproduced: The example gave `[a\, b]` instead of `[a|b, normal]`, silently dropping `normal`. `splitCells()` now splits only on unescaped pipes and unescapes `\|` in the content.

While verifying I found four more issues in the same delimiter logic:

- `-{3,}` rejected `|-|-|` (GFM needs only one dash)
- single-column tables were rejected
- a bare `---` could be misread as a delimiter row (now requires a pipe)
- __streaming `| :-: | :` flipped the block type table→prose→table and rebuilt the whole message — so aligned tables still flickered, the very thing this PR fixes__

These mattered for flicker too, since a rejected table falls back to the `TableRowSpan` path being replaced here.

One focused test for the escaped pipe as suggested, plus one per finding above (17 total, all passing). Alignment styling untouched. Branch squashed and rebased onto master.

wangzhaode (approved):

PR #4882, opened 2026-09-16
[CPU:Perf] Give each bilinear lane a whole C4 pixel

typer-J:
This is a follow-up to #4874, which landed the RVV C4 bilinear sampler with `vl` pinned to 4. It keeps that kernel's interface and its scalar path, and changes only how the vector lanes are assigned.

## What was wrong with the merged kernel

`vl` is 4 because there are four channels, and the four lanes *are* those channels — so one iteration covers exactly **one output pixel**, no matter how wide the hardware is. At VLEN=128 the `e8mf2` group has 8 lanes and half of them sit idle.

The load pattern that follows is the real cost. A lane only wants **one byte**, but an indexed load always moves a whole **32-bit element**, so three of every four fetched bytes are fetched and then discarded. The vector is doing three quarters of its memory traffic for nothing.

## What this changes

**Each lane now carries one whole C4 pixel**, i.e. one 32-bit word:

| | merged (`vl=4`) | this |
|---|---|---|
| lane means | one channel of a pixel | one whole pixel |
| pixels per iteration | **1** | **vl** (16 at VLEN=128, `e32m4`) |
| bytes per lane per fetch | 1 of 4 used | 4 of 4 used |
| addressing | 4 unit-stride `vle8` | 4 indexed `vluxei32` |

Because a lane consumes exactly the 32-bit element an indexed load moves anyway, **nothing is fetched and thrown away**. That is the whole point: the expensive addressing mode finally pays for itself.

The cost moves to the unpack — a word holds all four channels, so each channel is shifted out, the blend runs once per channel (reusing the per-pixel weights, which are the same for all four channels), and the results are shifted back into a word for a single store.

## Why this beats the merged kernel, not just the scalar

The measured claim is deliberately narrow: **the same addressing cost, three fewer quarters of wasted traffic.**

- At `vl=4` with `e8mf2` the four lanes of a unit-stride load fetch 4 bytes that are all used, but that is 4 bytes per **whole pixel**, and it is a separate load per neighbour — 16 bytes fetched per pixel across the four neighbours, plus 4 `vle8` and 4 `vzext_vf4` widenings.
- Here the four `vluxei32` each fetch a full word per lane and every byte is used, so the same 16 bytes per pixel arrive in 4 indexed loads whose cost is amortised over `vl` pixels instead of one.

Indexed addressing is genuinely more expensive per element on this core — a dedicated probe (`gather_probe.cpp`, same data, same order, only the addressing mode changed) measures `vle32` at 0.1651 ns/element against `vluxei32` at 0.4452–0.5440 ns/element, i.e. **2.7–3.3x**. The kernel wins because that higher unit price buys a fully-used element and 16 pixels of amortisation.

An intermediate design that also breaks the 4-lane limit — lanes = (pixel, channel) pairs — was measured and **lost to the merged kernel** (0.56–0.75x). It has *fewer* vector instructions per pixel than the merged kernel (16.8 vs 24.0) yet is ~34% slower, which is the clearest evidence that the bottleneck here is memory traffic, not ALU work. Breaking the lane limit is not the goal; using every fetched byte is.

## Verification

Board: riscv64 (VLEN=128), openEuler 24.03 SP3, g++ 14.3.1, Ninja, 32 cores, `taskset -c 8`.

**The kernel is really dispatched to.** Rather than calling `MNNSamplerC4Bilinear_RVV` directly, the benchmark links the real `libMNN.a`, calls `MNNGetCoreFunctions()`, and reads the sampler out of the slot the pipeline uses:

```
supportRVV                     : 1
core->MNNSamplerC4Bilinear     : 0x5b7f2
MNNSamplerC4Bilinear_RVV       : 0x5b7f2
slot == RVV kernel             : YES
```

The two pointers are bit-identical, so the registered slot *is* this kernel. If the registration were missing, the same run would report `NO` and the speedup below would collapse to 1.0x.

Registered test (`MNN_BUILD_TEST=ON`, `MNN_USE_RVV=ON`, static, Ninja, Release), rerun on the board with the follow-up safety patch applied:

```
RVV image-process: 54 cases passed (supportRVV=1)
```

The test drives the function table rather than the symbol and asserts `core->MNNSamplerC4Bilinear == MNNSamplerC4Bilinear_RVV`, so it exercises the registration. The original board run covered 17 output lengths x 3 point sets plus the rounding-boundary case. The follow-up safety patch adds an actual `aligned storage + 1 byte` source case through the registered function slot, plus allocation-free checks immediately below and above the complete `uint32_t` indexed-offset boundary. `-DMNN_USE_RVV=OFF` also configures, builds and runs cleanly. Regression: `backend/cpu/rvv` 3/3, `cv` 20/20.

**Board rerun of the follow-up commit** (riscv64, VLEN=128, openEuler 24.03 SP3, GCC 14.3.1, `taskset -c 8`). The three changed files were synced onto the board and the existing static build was rebuilt (`cmake --build . -j 32`, 49 targets, 37 s, exit 0), so these numbers come from the guarded kernel itself, not from the pre-patch binary:

| run | result |
|---|---|
| `backend/cpu/rvv/image_process` (`MNN_USE_RVV=ON`) | **54 cases passed** (`supportRVV=1`) |
| `backend/cpu/rvv` (full suite, `ON`) | **3/3 passed** |
| `cv` (`ON`) | **20/20 passed** |
| `backend/cpu/rvv/image_process` (`MNN_USE_RVV=OFF`) | passed (scalar registration path) |
| `cv` (`MNN_USE_RVV=OFF`) | **20/20 passed** |

The 54 matches the code: 17 output lengths x 3 point sets = 51, plus the rounding-boundary, unaligned-source and addressing-boundary cases = 54. The unaligned-source case is the one the guard exists for, and it passes through the registered function slot.

**Performance** (ns per output pixel, single core, fresh process, 5 rounds with arm order rotated, min over rounds; image 1920x1080, `CACHE_SIZE`=512 tile). Both vector arms come from the real library:

| geometry | generic scalar | merged `vl=4` | **this** | vs generic | **vs merged** |
|---|---|---|---|---|---|
| identity dx=1 | 39.26 | 15.79 | **11.40** | **3.45x** | **1.39x** |
| scale 2x dx=0.5 | 39.29 | 15.80 | **12.03** | **3.27x** | **1.31x** |
| skew dy=0.75 | 39.35 | 15.85 | **12.09** | **3.26x** | **1.31x** |
| fractional 0.3 | 39.33 | 15.80 | **12.87** | **3.06x** | **1.23x** |

**Output equivalence.** Compared directly against the generic body in the same process, same input: identical on identity, 2x and skew; the fractional geometry differs on 10 of 8.3M bytes, each by one grey level, which is the documented FP64/FP32 divergence in the `c10` term (`yF * (1.0 - xF) * c10` keeps `1.0 - xF` in double in the generic body). The sentinel margin after the written range stays byte-exact, so a write past the requested range would fail rather than being absorbed.

**Scalar fallback.** The vector path now requires both the source base and row stride to be 4-byte aligned, because it casts the byte pointer to `uint32_t*` and issues 32-bit indexed loads. It also checks the complete maximum byte offset, `(ih - 1) * yStride + 4 * (iw - 1)`, with overflow-safe division before entering the vector path. An unaligned base, a partial-pixel stride, a row-plus-column offset outside `uint32_t`, or a core wider than the lane buffer takes the scalar path instead of risking an alignment trap or wrapped address. This is also strictly better than the merged kernel at low VLEN: that one pins `vl` to 4 in `e8mf2` and must fall back when the hardware offers fewer than 4 lanes, whereas `e32m4` still yields 4 lanes at VLEN=32, so this version stays vectorised on narrow cores where the merged one does not.

## Module

CPU

## Type

- [ ] Feature
- [x] Perf
- [ ] Bugfix
- [ ] Refact
- [ ] Style
- [ ] Doc
- [ ] Test
- [ ] Chore


typer-J:
Relationship to #4874: that PR has already landed as `60e10f81`, so this one is filed fresh against current `master` rather than pushed to the merged branch. It touches a single file — the kernel #4874 added — and keeps its interface and scalar fallback unchanged, so it applies cleanly on top and is independent of the other RVV work.

The measured baseline labelled "merged" in the description is the `vl=4` kernel exactly as it landed, read out of the same function-table slot in the same process.

wangzhaode:
Thanks for the follow-up optimization and the detailed correctness and performance results! Processing a whole C4 pixel per lane looks like a useful improvement.

Before merging, could you please add a source-pointer alignment check to the vector-path guard? The new path casts `source` to `const uint32_t*` and uses 32-bit indexed loads. Checking `(yStride & 3) == 0` does not ensure that `source` itself is 4-byte aligned, and the ImageProcess byte-pointer interface does not guarantee that alignment. On RVV implementations that do not support misaligned vector loads, this can raise an exception. Falling back to the existing scalar path when `(reinterpret_cast<uintptr_t>(source) & 3) != 0` should be sufficient.

While updating the guard, please also cover the complete indexed offset: the current check bounds `(ih - 1) * yStride`, but the subsequent addition of `4 * x` can still overflow a 32-bit offset. An overflow-safe check of the maximum row-plus-column offset before entering the vector path would address this.

A small unaligned-source case (for example, an aligned allocation plus one byte) and a focused check of the offset boundary are enough; there is no need for a broad new benchmark suite or a multi-gigabyte test allocation. Thanks again!

typer-J:
Thanks for catching both cases. I addressed them in 79f9677e:/n/n- require the source base and row stride to be 4-byte aligned before entering the / path; otherwise use the existing scalar fallback;\n- validate the complete maximum byte offset, , using overflow-safe division rather than checking only the row term;\n- add an actual aligned-allocation-plus-one-byte source case through the registered function-table entry, plus allocation-free boundary checks for the largest valid offset, row-plus-column overflow, column overflow, unaligned stride, and empty dimensions.\n\nLocal validation completed: , clang-format, an  C++11 cross-syntax check of the RVV translation unit, a syntax check of the test translation unit, and direct execution of the address-boundary helper cases all pass. I could not rerun the full RISC-V board binary from this environment because the previously used board SSH credentials are unavailable here, so I have stated that limitation explicitly in the PR description rather than claiming a board result for this follow-up commit.

typer-J:
Reran this on the actual board, so the earlier caveat in the PR description is gone.

The three changed files were synced onto the board (md5-verified against the commit blobs) and the existing static build was rebuilt in place (`cmake --build . -j 32`, 49 targets, 37 s, exit 0). So the numbers below come from the guarded kernel, not from the pre-patch binary.

Board: riscv64, VLEN=128, openEuler 24.03 SP3, GCC 14.3.1, 32 cores, `taskset -c 8`.

```
RVV image-process: 54 cases passed (supportRVV=1)
```

| run | result |
|---|---|
| `backend/cpu/rvv/image_process` (`MNN_USE_RVV=ON`) | 54 cases passed (`supportRVV=1`) |
| `backend/cpu/rvv` full suite (`ON`) | 3/3 passed |
| `cv` (`ON`) | 20/20 passed |
| `backend/cpu/rvv/image_process` (`MNN_USE_RVV=OFF`) | passed, scalar registration path |
| `cv` (`MNN_USE_RVV=OFF`) | 20/20 passed |

The 54 lines up with the code: 17 output lengths x 3 point sets = 51, plus the rounding-boundary, unaligned-source and addressing-boundary cases = 54. The unaligned-source case is exactly the one the guard exists for, and it goes through the registered function slot rather than calling the symbol directly.

Both cases you raised are covered by that guard:

- the source base and row stride must be 4-byte aligned before the `uint32_t` / `vluxei32` path is taken, otherwise it falls back to the scalar kernel;
- the complete maximum byte offset `(ih - 1) * yStride + 4 * (iw - 1)` is validated with overflow-safe division, so the row term alone is no longer the only thing checked.

Thanks again for catching these — the alignment one in particular was a real hole.

typer-J:
Thanks for catching both cases. I addressed them in 79f9677e and reran the updated code on the actual board.

The patch now:

- requires both the source base and row stride to be 4-byte aligned before entering the `uint32_t` / `vluxei32` path; otherwise it uses the existing scalar fallback;
- validates the complete maximum byte offset, `(ih - 1) * yStride + 4 * (iw - 1)`, using overflow-safe division rather than checking only the row term;
- adds an actual aligned-allocation-plus-one-byte source case through the registered function-table entry, together with allocation-free checks for the largest valid offset, row-plus-column overflow, column overflow, unaligned stride, and empty dimensions.

For the board rerun, the three changed files were synced and MD5-verified against the 79f9677e commit blobs. The existing static build was then rebuilt in place (`cmake --build . -j 32`, 49 targets, 37 s, exit 0), so the results below come from the guarded kernel rather than the pre-patch binary.

Since this follow-up does not change the RVV interpolation, unpacking, weight computation, or rounding logic, I did not repeat the earlier large-scale equivalence benchmark over approximately 8.3 million output bytes. Instead, I reran the 54-case registered-path correctness test described below, together with the related regression suites.

Board: riscv64, VLEN=128, openEuler 24.03 SP3, GCC 14.3.1, 32 cores, `taskset -c 8`.

| run | result |
|---|---|
| `backend/cpu/rvv/image_process` (`MNN_USE_RVV=ON`) | 54 cases passed (`supportRVV=1`) |
| `backend/cpu/rvv` full suite (`ON`) | 3/3 passed |
| `cv` (`ON`) | 20/20 passed |
| `backend/cpu/rvv/image_process` (`MNN_USE_RVV=OFF`) | passed through the scalar registration path |
| `cv` (`MNN_USE_RVV=OFF`) | 20/20 passed |

The 54 cases consist of 17 output lengths x 3 point sets = 51, plus the rounding-boundary, unaligned-source, and addressing-boundary cases. The unaligned-source case goes through the registered function slot rather than calling the RVV symbol directly.

Both issues you raised are therefore covered by the guard and by focused regression tests. Thanks again for catching them — the missing source-alignment check in particular was a real correctness hole.

wangzhaode (approved):

PR #4894, opened 2026-09-19
[LLM:Bugfix] Refresh PLE for successive Omni text requests

### LINKED ISSUE #4890: [Bug] Omni reuses stale PLE input across successive text requests
## Summary

With Gemma 4 E2B, the first text request completes normally, but subsequent requests after resetting the conversation on the same inference instance can produce unrelated or repetitive output until the token limit.

The text-only embedding path appears to reuse PLE input left over from the previous decode step.

## Platform and version

- Device: HUAWEI MatePad Edge, Kirin X90, 32 GB RAM.
- OS: HarmonyOS, arm64-v8a.
- Inference backend: CPU.
- MNN revision tested: `2edeef91b425e98a93707840b6fffdd97980bdbe`.

## Model and reproduction

Model: `MNN/gemma-4-E2B-it-MNN`, using the same model files for all comparisons.

Reproduced through our application using MNN’s C++ LLM interface:

1. Load the model and send a text-only prompt.
2. Wait for generation to finish.
3. Reset the conversation without unloading the model.
4. Send another text prompt and repeat.

The first request completed normally; the following four produced abnormal output and reached the 512-token limit. A standalone `llm_demo` reproduction has not yet been verified.

## Runtime logs and diagnosis

Diagnostic logging captured a new **109-token input** entering the embedding path while the cached `mPleInput` had a **sequence length of 1**.

In [`Omni::embedding()`](https://github.com/alibaba/MNN/blob/2edeef91b425e98a93707840b6fffdd97980bdbe/transformers/llm/engine/src/omni.cpp#L2049), the text-only branch calls `Llm::embedding()` without clearing the previous `mPleInput`. The [base implementation](https://github.com/alibaba/MNN/blob/2edeef91b425e98a93707840b6fffdd97980bdbe/transformers/llm/engine/src/llm.cpp#L1557) only recomputes PLE when:

```cpp
mPleEmbedding && (!mPleInput.get() || seq_len == 1)
```

A subsequent multi-token prompt therefore bypasses recomputation when PLE remains non-null after the previous single-token decode step.

Preserving precomputed PLE is needed for multimodal prefill, but this condition does not distinguish current multimodal data from stale data belonging to an earlier request.

## Locally verified fix

Inside Omni’s text-only branch, before calling `Llm::embedding()`:

```cpp
if (mPleEmbedding && input_ids.size() > 1) {
    mPleInput = nullptr;
}
```

This allows PLE to be recomputed for the current text input while leaving the multimodal branch unchanged.

With the same application code and model files, changing only the MNN library:

- Five fresh text requests completed normally after the fix.
- A two-turn recall test and image/text alternation also passed.

These results support this specific stale-PLE diagnosis; they do not establish that other Gemma output issues share the same cause.

ZedingZhang:
## Description

Fixes #4890.

Gemma 4 Omni instances can retain the one-token PLE input produced by the final decode step. A later multi-token text prefill then sees a non-null `mPleInput`, so `Llm::embedding()` keeps the stale tensor instead of rebuilding PLE for the new request.

This change:

- clears `mPleInput` before a multi-token prefill only when Omni is on the text-only path and PLE is enabled; and
- leaves the multimodal path unchanged so its full-input precomputed PLE is preserved.

## Module

LLM / Omni (Gemma 4 PLE)

## Type

- [ ] Feature
- [x] Bugfix
- [ ] Perf
- [ ] Refact
- [ ] Style
- [ ] Doc
- [ ] Test
- [ ] Chore

## Testing

- Validated with a temporary `test_omni_ple_lifecycle` executable built using `MNN_LLM_BUILD_TEST=ON`; the test source and CMake target were removed per maintainer review.
- Negative control: after reverting only the fix, the temporary test fails with `Text prefill reused the stale one-token PLE input`.
- Fixed implementation: the same test passes and observes the PLE sequence length changing from 1 to 3.
- A/B CI run: https://github.com/ZedingZhang/MNN/actions/runs/35440540720
- Fork platform CI passed on [Android](https://github.com/ZedingZhang/MNN/actions/runs/35440470832), [iOS](https://github.com/ZedingZhang/MNN/actions/runs/35440470831), [Linux](https://github.com/ZedingZhang/MNN/actions/runs/35440470856), [macOS](https://github.com/ZedingZhang/MNN/actions/runs/35440470869), and [Windows](https://github.com/ZedingZhang/MNN/actions/runs/35440470876).

## Checklist


wangzhaode:
Thanks for the focused PLE lifecycle fix and the A/B validation!

As with #4888, please remove `transformers/llm/engine/test/test_omni_ple_lifecycle.cpp` and its corresponding executable, include-directory, and link entries in `transformers/llm/engine/CMakeLists.txt`. This test covers a very narrow scenario, and we would prefer not to retain a separate test executable for it. The validation already documented in the PR is sufficient for this change; there is no need to move or expand the test.

Please keep the core PLE refresh fix in `Omni::embedding()` unchanged. Thanks again!

ZedingZhang:
I’ve removed `test_omni_ple_lifecycle.cpp` and its corresponding executable, include-directory, and link entries from `transformers/llm/engine/CMakeLists.txt`. No replacement test was added.

The core PLE refresh fix in `Omni::embedding()` remains unchanged. I updated the PR description to clarify that the temporary test was used only for A/B validation and is no longer part of the change. The PR now contains only the core `omni.cpp` fix. Thanks!

wangzhaode (approved):

